PR #13, opened 2021-05-11
feat: add support for volume mounts

victorlcm:
Hi everyone,

I'm currently trying to use the k6 operator to implement a load testing solution in the company I work. We have some pretty aggressive scenarios with some big files to be loaded in the init phase to be used by all VUs. Due to these big files, I got an error when trying to create the scripts as config maps, because they cannot exceed 3mb in size. So, to solve this I've modified the operator a little bit to also accept a persistent volume claim. With the claim, it will not mount volume using config maps, but will mount using the claim name. With this, we can add the code anywhere it's supported by Kubernetes and we'll not have issues with file sizes. I've also increased the manager requests/limits a little bit as the pod was getting a lot of OOMKill from the cluster.

Please let me know whether I can improve anything and I'll do it asap!


victorlcm:
Hey @simskij 

Thanks! Sure, I'll try my best to contribute with both suggestions, ideas and also some code! hahaha
This tool is so good and is helping a lot! distributing is always hard and this operator made it so much better

victorlcm:
hey @simskij,

I've just resolved the conflicts in this PR, did you have the chance to test it?

config/manager/manager.yaml
```diff
@@ -33,8 +33,8 @@ spec:
         resources:
           limits:
             cpu: 100m
-            memory: 30Mi
+            memory: 60Mi
```
simskij, line 36: Let's bump this some more. As it's the burst limit, it doesn't really make sense to keep it too low, as you pointed out.
```suggestion
            memory: 100Mi
```

pkg/resources/jobs/runner.go
```diff
@@ -72,7 +72,7 @@ func NewRunnerJob(k *v1alpha1.K6, index int) (*batchv1.Job, error) {
 						Ports: ports,
 					}},
 					TerminationGracePeriodSeconds: &zero,
-					Volumes:                       newVolumeSpec(k.Spec.Script),
+					Volumes:                       newVolumeSpec(k.Spec.Script, k.Spec.VolumeClaim),
```
simskij, line 75: Maybe it would be preferable to pass the whole Spec instead of adding an additional param?

api/v1alpha1/k6_types.go
```diff
@@ -27,6 +27,7 @@ type K6Spec struct {
 	Arguments   string                 `json:"arguments,omitempty"`
 	Image       string                 `json:"image,omitempty"`
 	Ports       []corev1.ContainerPort `json:"ports,omitempty"`
+	VolumeClaim string                 `json:"volumeClaim,omitempty"`
```
simskij, line 30: See line 24. If it's now possible to pick between using a volume and a config map, the script prop should be made optional as well.

simskij (changes requested):
LGTM! Some minor suggestions

simskij:
Over time, I think it would be desirable to make the usage like this (but I'm definitely not expecting you to do that change as part of this PR, just want to give you an idea of where I think we should be heading 😅 ):

```yaml
spec:
  script:
    volumeClaim:
      name: k6-test-files
      file: /a.test.bundle.js
    # or...
    configMap:
      name: k6-test-files-configmap
      file: test.js
```

victorlcm:
> Over time, I think it would be desirable to make the CRD look like this (but I'm definitely not expecting you to do that change as part of this PR, just want to give you an idea of where I think we should be heading 😅 ):
> 
> ```yaml
> spec:
>   script:
>     volumeClaim:
>       name: k6-test-files
>       file: /a.test.bundle.js
>     # or...
>     configMap:
>       name: k6-test-files-configmap
>       file: test.js
> ```

Hey @simskij 👋 

I really liked this change so I've already implemented and pushed into this PR. Could you please review it again? I'm new to Golang, so any inputs are appreciated 😄

pkg/resources/jobs/runner.go
```diff
@@ -126,3 +151,35 @@ func newLabels(name string) map[string]string {
...
+		if spec.Script.ConfigMap.File != "" {
+			s.File = spec.Script.ConfigMap.File
+		}
+
+		s.Type = "ConfigMap"
+		return s, nil
+	}
+
+	return nil, errors.New("ConfigMap or VolumeClaim not provided in script definition")
+}
```
simskij, line 185:
```suggestion
func newScript(spec v1alpha1.K6Spec) (*Script, error) {
      s := &Script{}
      s.File = "test.js"
        
	if spec.Script.VolumeClaim.Name != "" {
		s.Name = spec.Script.VolumeClaim.Name		
		if spec.Script.VolumeClaim.File != "" {
			s.File = spec.Script.VolumeClaim.File
		}

		s.Type = "VolumeClaim"
		return s, nil
	}

	if spec.Script.ConfigMap.Name != "" {
		s.Name = spec.Script.ConfigMap.Name

		if spec.Script.ConfigMap.File != "" {
			s.File = spec.Script.ConfigMap.File
		}

		s.Type = "ConfigMap"
		return s, nil
	}

	return nil, errors.New("ConfigMap or VolumeClaim not provided in script definition")
}
```

pkg/resources/jobs/runner.go
```diff
@@ -28,13 +36,19 @@ func NewRunnerJob(k *v1alpha1.K6, index int) (*batchv1.Job, error) {
...
+	}
+
 	if k.Spec.Arguments != "" {
 		args := strings.Split(k.Spec.Arguments, " ")
 		command = append(command, args...)
 	}
 	command = append(
 		command,
-		"/test/test.js",
+		"/test/"+script.File,
```
simskij, line 51:
```suggestion
		fmt.Sprintf("/test/%s", script.File),
```

simskij (changes requested):
Two more minor changes, then we're good to go! 🎉

victorlcm:
thx for the comments!

simskij:
Thanks for the contribution! 🎉

Issue #15, opened 2021-05-24
Integration with Prometheus via statsd exporter


jmdacruz:
I'm trying to figure out how to get the k6 metrics into Prometheus. One alternative is to deploy https://github.com/prometheus/statsd_exporter and have k6 send the metrics there with the `--out statsd` argument  (https://k6.io/docs/results-visualization/statsd/). Is there a way to inject the required environment variables (e.g., `K6_STATSD_ADDR`) with the operator?

jmdacruz:
thanks @dgzlopes, I completely missed the `-e` argument! And yes, it would be ideal to be able to replace the image on the operator. I'll see if I can put together a PR for this.

dgzlopes:
Now that #17 has been merged, we can close this :)

btw, thanks for raising the issue!

mikaelkundert:
I did not get `-e` argument working.

I got following errors:
```
time="2021-06-04T14:51:39Z" level=error msg="Couldn't flush a batch" error="write udp 127.0.0.1:36848->127.0.0.1:8125: write: connection refused" output=statsd
```

While having following manifest:
```
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6-sample
spec:
  parallelism: 4
  script: crocodile-stress-test
  arguments: -e K6_STATSD_ADDR=graphite-1622816495:8125 --out statsd
  ```
  
This popped up job with commands:
```
k6 run --quiet --execution-segment=0:1/4 --execution-segment-sequence=0,1/4,2/4,3/4,1 -e K6_STATSD_ADDR=graphite-1622816495:8125 --out statsd /test/test.js --address=0.0.0.0:6565 --paused

/\ |‾‾| /‾‾/ /‾‾/
/\ / \ | |/ / / /
/ \/ \ | ( / ‾‾\
/ \ | |\ \ | (‾) |
/ __________ \ |__| \__\ \_____/ .io
execution: local
script: /test/test.js
output: statsd (localhost:8125)
```

But this doesn't work locally either. Unless I run the command as follow:
```
$ K6_STATSD_ADDR=graphite-1622816495:8125 k6 run --quiet --execution-segment=0:1/4 --execution-segment-sequence=0,1/4,2/4,3/4,1 -e K6_STATSD_ADDR=graphite-1622816495:8125 --out statsd ./test.js --address=0.0.0.0:6565 --paused

          /\      |‾‾| /‾‾/   /‾‾/   
[...]

PR #16, opened 2021-05-24
Add support for job environment

jmdacruz:
Simply add the `Env` property to the CRD, and pass that to the K6 job spec. This should enable uses cases such as #15


dgzlopes:
Thanks for opening the PR :muscle: 

As I showcased on #15, this can be accomplished by using the `-e` flag in k6, so I think, that it's a bit redundant to add the Env property to the CRD. 

Thanks anyway!

PR #17, opened 2021-05-26
Support for overriding the k6 image

jmdacruz:
Support for overriding the k6 image. Suggested by @dgzlopes  in https://github.com/k6io/operator/issues/15#issuecomment-848648427

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,48 @@ After completing a test run, you need to clean up the test jobs created. This is
 $ kubectl delete -f /path/to/your/k6-resource.yml
 ```
 
+### Using extensions
+By default, the operator will use `loadimpact/k6:latest` as the container image for the test jobs. If you want to use
+extensions built with [xk6](https://github.com/k6io/xk6) you'll need to create your own image and override the `image`
+property on the `K6` kubernetes resource. For example, the following Dockerfile can be used to create a container
+image using github.com/szkiba/xk6-prometheus as an extension:
+
+
+```Dockerfile
+# Build the k6 binary with the extension
+FROM golang:1.16.4-buster as builder
+
+RUN go install github.com/k6io/xk6/cmd/xk6@latest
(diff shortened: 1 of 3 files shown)


dgzlopes (approved):
LGTM! Thanks :)

jmdacruz:
While trying to test this approach, I realize that we would also need the ability to expose extra ports on the container (so that Prometheus can scrape the metrics on port `5656` in this case). I could add this capability on this PR too

jmdacruz:
@dgzlopes I took a stab at supporting extra ports on the CR, so that we can expose other ports for custom operator images

majnos:
Would be nice if you could also update readme how to make prometheus work with k6-operator. Just some bullet points.

jmdacruz:
@majnos I just added updated the readme with info on how use extensions, with the Prometheus extension as an example

majnos:
something is wrong ... not sure what

time="2021-05-28T11:58:56Z" level=error msg="invalid output type 'prometheus', available types are: cloud, csv, datadog, influxdb, json, kafka, statsd"

jmdacruz:
@majnos note that the error happens on your last stage, when you are trying to copy something that the "builder" image probably does not have (the `/workspace/manager` folder).

jmdacruz:
@dgzlopes what else do we need to merge this PR?

dgzlopes:
Nothing more :) 

Thanks for adding the extra docs! Merging.

jmdacruz:
thanks @dgzlopes! One last thing: how do we trigger a new build for the operator image so that Docker Hub is updated?

Juandavi1:
Hi guys, this still dont works :(

dgzlopes:
Sorry for the late response! @simskij updated the Docker Image. Everything should work now.

dgzlopes:
We should add #20 😄

Juandavi1:
It works !! Thaks guys

PR #24, opened 2021-06-05
feat: Add support for custom annotations and labels for runners and starter

### LINKED ISSUE #23: Support for Custom Annotations/Labels on Starter and Runner
Currently if running k6 in an istio environment the starter fails because the curl command is sent too quickly. The istio/envoy proxy hasn't started at the point and there is no retry built in.

While adding a retry is probably a reasonable option (one for a separate issue), being able to add custom annotations like: 

```
annotations:
  proxy.istio.io/config: '{ "holdApplicationUntilProxyStarts": true }'
```

or:

```
sidecar.istio.io/inject: "false"
```

would be really useful.

I'm going to take a crack at this, but broadly I think it can be thought of as creating the struct/spec for a starter and a runner object, and then accepting annotations/labels for each of them solves the issue

KnechtionsCoding:
Resolves #23 and allows for adding of custom annotations and labels for runners and the starter.

Reasonably straight forward PR.

I did have a question about the imports however, it appears there are duplicate imports in runner.go and starter.go (k6 is imported in both of them but doesn't need to be, and is done so under 2 different names. As is core api from k8s). I can open another PR to create a helper.go that has newLabels, and the imports in it if desired. I found it a little confusing for a moment and is probably unnecessary.


simskij:
Mistakenly merged #26 prior to this one, but since all commits are present there as well I'll just go ahead and close this one. Thanks!

PR #27, opened 2021-06-06
Refactor/standardize names

KnechtionsCoding:
Builds off of #24 . When I was working on #24 I noticed three things: 

- that there was 2 different names for the k6 object in the jobs package
- the name of the import corev1 api was different
- that there wasn't a clean way to share code between the files.

This PR fixes all three.

- Changes the k6 object that is used as an arg from k to k6 everywhere in runner.go to match starter.go (this made the most sense because `k` is rather non-descriptive
- Changes the name of the import from v1 to corev1 in starter.go to match runner.go (makes more sense as v1 could refer to any number of things; in the same file three other apis with v1 are imported)
- Creates helpers.go and puts the `newLabels` func in it to make it clear where it comes from because it is now used in both runner.go and starter.go rather than just runner.go


simskij (approved):

simskij:
Thanks for the contribution!

Issue #20, opened 2021-06-02
Setup CI for development and production builds


simskij:
Using GitHub Actions.

KnechtionsCoding:
@simskij Do we know what CI Operations are wanted? Test, bundle, etc? Happy to write some github actions to help out

dgzlopes:
I think that the biggest necessity right now (apart from #4) is building/pushing new Docker images on each release.

Right now this process is manual, and adds a bit of friction.

simskij:
@KnechtionsCoding that would be great! Yeah, testing, bundling and pushing to ghcr would probably do. I'd be happy to set up any secrets needed (if any, not sure since it's all GH).

KnechtionsCoding:
@simskij @dgzlopes Is that the `make bundle` command? I posted #28 when I tried to run it locally to see what the output was. 

I'll work on figuring out a GH action for the rest of these in the mean time.

immavalls:
Adding the bundle generation on releases would indeed help users like https://community.k6.io/t/make-deploy-issue-from-controller-gen/5871/. We see some struggles building the operator when following https://k6.io/blog/running-distributed-tests-on-k8s/

PR #30, opened 2021-06-07
Feat/test-github-action

KnechtionsCoding:
### Summary

Adds two github actions:
- test (on matric of versions)
- publish to ghcr on release publish events

Modifies test setup:
- uses newer version of setupenv-test and kubebuilder to allow for in ci testing

Docs:
- updates docs to cover new commands in makefile

### Details:

When running in github, we obviously don't have etcd or other tooling installed by default. The script that was being used to download all of the env setup wasn't working properly in github (I have some failed actions demonstrating that).
In April of this year, controller-runtime switched to a new method: https://github.com/kubernetes-sigs/controller-runtime/commit/782807d8d4fb287e86f6c91862529574ac42cc0a#diff-051375546db9782e3debc25e0241edf1d5e5e2ec0f183dd8634ca5b2c8968bb8

I mimic'd that method with a few modifications to match what was currently happening. I left everything as a makefile script for now, but we may want to move the test-setup command to a bash script at some point to make it easier.
You'll note the addition of an ENV variable, that's based on: https://github.com/kubernetes-sigs/kubebuilder/issues/326

I've currently dropped the kube-builder tools into /tmp/kube-builder but they could theoretically be anywhere. This is really only relevant to people working on the operator.

If there is a different method wanted, then I can explore that route.

I've also configured it so it can run a matrix of tests. This will allow us to test the operator on a variety of linux platforms, with a variety of golang versions, and a variety of k8s versions. I've only included one of each for now for demonstration,
but we can obviously add more.

works towards #20 

Successful example here: https://github.com/KnechtionsCoding/operator/runs/2765262646

### Next Steps

@simskij will probably need to follow: https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#upgrading-a-workflow-that-accesses-ghcrio to allow the actions/workflow to publish to ghcr


.github/workflows/push.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - name: "Build:checkout"
+        uses: actions/checkout@v2
+      - name: "Build:dockerimage"
+        uses: docker/build-push-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+          repository: k6io/operator
+          tags: latest
```
dgzlopes, line 22: I think we can pass a list of tags here. That way, we don't have to repeat the `Build:dockerimage` block.

dgzlopes (approved):
Thanks! Left one small comment :)

simskij:
Added the permissions needed to ghcr.io. Once you've sorted out @dgzlopes suggestion, I'm happy to merge it. :)

PR #31, opened 2021-06-16
chore: fix duplicate declaration because of merge conflict

KnechtionsCoding:
When the other PRs from me where merged there was a duplicate declaration this fixes it and should now be good to go


simskij:
Yeah, that was probably me. Thanks for catching it!

Issue #32, opened 2021-06-18
Allow the K6 CRD to set environment variables for the runners.


simskij:
> I did not get `-e` argument working.
> 
> I got following errors:
> ```
> time="2021-06-04T14:51:39Z" level=error msg="Couldn't flush a batch" error="write udp 127.0.0.1:36848->127.0.0.1:8125: write: connection refused" output=statsd
> ```
> 
> While having following manifest:
> ```
> apiVersion: k6.io/v1alpha1
> kind: K6
> metadata:
>   name: k6-sample
> spec:
>   parallelism: 4
>   script: crocodile-stress-test
>   arguments: -e K6_STATSD_ADDR=graphite-1622816495:8125 --out statsd
>   ```
>   
> This popped up job with commands:
> ```
> k6 run --quiet --execution-segment=0:1/4 --execution-segment- sequence=0,1/4,2/4,3/4,1 -e K6_STATSD_ADDR=graphite-1622816495:8125 --out statsd /test/test.js --address=0.0.0.0:6565 --paused
> 
> /\ |‾‾| /‾‾/ /‾‾/
> /\ / \ | |/ / / /
> / \/ \ | ( / ‾‾\
> / \ | |\ \ | (‾) |
> / __________ \ |__| \__\ \_____/ .io
> execution: local
> script: /test/test.js
> output: statsd (localhost:8125)
> ```
> 
> But this doesn't work locally either. Unless I run the command as follow:
> ```
> $ K6_STATSD_ADDR=graphite-1622816495:8125 k6 run --quiet --execution-segment=0:1/4 --execution-segment-sequence=0,1/4,2/4,3/4,1 -e K6_STATSD_ADDR=graphite-1622816495:8125 --out statsd ./test.js --address=0.0.0.0:6565 --paused
> 
>           /\      |‾‾| /‾‾/   /‾‾/   
>      /\  /  \     |  |/  /   /  /    
>     /  \/    \    |     (   /   ‾‾\  
>    /          \   |  |\  \ |  (‾)  | 
>   / __________ \  |__| \__\ \_____/ .io
> 
> ERRO[0000] The moduleSpecifier "./test.js" couldn't be found on local disk. Make sure that you've specified the right path to the file. If you're running k6 using the Docker image make sure you have mounted the local directory (-v /local/path/:/inside/docker/path) containing your script and modules so that they're accessible by k6 from inside of the container, see https://k6.io/docs/using-k6/modules#using-local-modules-with-docker.
> ```

[...]

PR #33, opened 2021-06-24
Feat/env vars runner

KnechtionsCoding:
solves #32


KnechtionsCoding:
@simskij this solves #32 but not sure if we want to be able to provide env variables to the starter. If not, this works, if we do, then I probably need to add to the runner/start structs I created before for metadata.

Issue #8, opened 2021-02-19
Add nodeSelectors and/or nodeAffinity/anti-affinity


dgzlopes:
The Kubernetes scheduler can be constrained to place a pod on particular nodes using few different options. In your POD specification, there are many ways in which you declare a POD should be dedicated to specific nodes, one of them is nodeSelectors and another one is nodeAffinity/anti-affinity.

I would like to constrain where I can place k6 runners. This would open up a lot of fancy possibilities. 

For example: 

- Only use nodes from certain zones (and use Separate to spread the runners). 
- Run my test on some type of instance (e.g. M4). 
- Run my test on an isolated node pool of my cluster.

PR #34, opened 2021-06-24
Feat/pod struct

KnechtionsCoding:
Creates a PodSpec, migrates podMetadata under the PodSpec. It also adds the NodeSelector, AntiAffinity, and Env to the Runner. NodeSelector and AntiAffinity are also passed to the starter object (env isn't, but can be added easily).


KnechtionsCoding:
Closes #32 and #8

mikaelkundert:
This idea and solutions looks good.

I was slightly worried about using name `Pod` since it could cause some confusion with `corev1.Pod` or `corev1.PodSpec`. But after longer thought I think using `Pod` term is OK.

KnechtionsCoding:
@mikaelkundert the other option was something like `k6pod` but that seemed extra verbose and not super necessary

mikaelkundert:
> @mikaelkundert the other option was something like `k6pod` but that seemed extra verbose and not super necessary

Yeah I agree. I think `Pod` might be OK here. This term is used only within Go and not in Yamls, so contributors should be aware of the context pretty easily.

PR #35, opened 2021-07-02
Fix/service account

KnechtionsCoding:


mikaelkundert:
Worked out in my local testing. LGTM

PR #45, opened 2021-07-13
remove configmap/status resource from leader-election-role

### LINKED ISSUE #44: leader-election-role role refers to non-existent configmaps/status resource
The leader-election-role at https://github.com/k6io/operator/blob/main/config/rbac/leader_election_role.yaml#L19 refers to a resource configmaps/status, but as far as I know configmaps don't have a status (https://v1-20.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#configmap-v1-core). Perhaps I'm misunderstanding something though?

I removed that item from the rules array and the project appears to run just fine.

mycargus:
fixes #44

### DIFF
--- a/config/rbac/leader_election_role.yaml
+++ b/config/rbac/leader_election_role.yaml
@@ -19,14 +19,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - configmaps/status
-  verbs:
-  - get
-  - update
-  - patch
-- apiGroups:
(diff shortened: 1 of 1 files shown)


KnechtionsCoding:
This looks good to me. I've tested the same change and everything appears to run without issue.

dgzlopes (approved):

Issue #43, opened 2021-07-13
Bug: Two k6 Jobs called the same thing


KnechtionsCoding:
If two k6 jobs are called the same thing, even in different namespaces, the k6 controller gets confused and does something like:

```txt
2021-07-13T15:08:22.864Z        INFO    controllers.K6  Waiting for pods to get ready   {"k6": "big-brother/k6-test"}
2021-07-13T15:08:22.865Z        INFO    controllers.K6  5/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:08:27.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:08:32.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:08:37.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:08:42.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:08:47.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:08:52.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:08:57.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:09:02.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:09:07.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:09:12.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:09:17.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:09:22.866Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
2021-07-13T15:09:22.867Z        INFO    controllers.K6  9/4 pods ready  {"k6": "big-brother/k6-test"}
```

k6_start.go needs to be modified to poll for labels and namespaces as well to prevent issues

PR #46, opened 2021-07-14
fix: only search a specific namespace

KnechtionsCoding:
Currently two k6 tests named the same thing in two different namespaces are returned when searching, this breaks the start functionality.

Limit the search for an individual k6 job to the namespace the k6 job was created in.


KnechtionsCoding:
closes #43

simskij (approved):

PR #49, opened 2021-07-16
feat: add support for Service Account Name

knmsk:
Hi everyone,
Currently I'm trying to use the k6-operator to build a chart for load testing on the company I work, in our case we need to set a custom service account name to be used by the job/pod to enable the vault sidecar to inject the volumes on the pod.

Please let me know whether I can improve anything and I'll do it asap!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
 # Image URL to use all building/pushing image targets
 IMG ?= ghcr.io/k6io/operator:latest
+# Default dockerfile to build
+DOCKERFILE ?= "Dockerfile.controller"
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
@@ -98,7 +100,7 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build . -t ${IMG}
+	docker build . -t ${IMG} -f ${DOCKERFILE}
 
 # Push the docker image
 docker-push:
--- a/api/v1alpha1/k6_types.go
+++ b/api/v1alpha1/k6_types.go
@@ -37,14 +37,13 @@ type Pod struct {
 
 // K6Spec defines the desired state of K6
 type K6Spec struct {
-	Script             K6Script               `json:"script"`
-	Parallelism        int32                  `json:"parallelism"`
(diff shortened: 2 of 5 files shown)


KnechtionsCoding:
@knmsk Sorry, I added support for the serviceAccount in a different way, and missed this PR. I actually think your way might is better, and includes the automount option(which I've never seen before). 

What's the scenario in which you want to not automount the service token?

KnechtionsCoding:
But otherwise, if you rebase, and then move it serviceAccount from a top level item and keep yours where it is, I think we can get this merged. It's a good change.

knmsk:
@KnechtionsCoding in my case, we want to customize the serviceAccount to add some metadata tags like `app.kubernetes.io/managed-by` and `app.kubenetes.io/name` to have a better control of resources created using a Helm Chart.

KnechtionsCoding:
Seems reasonable to me.

knmsk:
@KnechtionsCoding i've made the rebase and added some changes:
1. If anyone did not send the `automountServiceAccountToken` property it's always gonna be `false` so i switched to string to have the default set to `true`.
2. I thought of maintaining the `serviceaccount` on the top level item, to have a clean method to add the serviceAccount to both the runner and starter without adding more one more line on the yaml... (If this changes does not make sense please let me know that i will remove it)
3. Also, added a dockerfile on the makefile config to allow build the controller.

**Obs**: Somehow i messed up my personal github account and my work account in this last commit :disappointed:

KnechtionsCoding (changes requested):
Can you make sure to remove serviceAccount from the top level k6 struct as well, seeing as it isn't being used here?

KnechtionsCoding:
@knmsk I originally added the service account to the top level. I think there are possibly cases where you want two different service accounts, and seeing as it's not super bad to have one extra line (and in line with the pod spec in general), I think this is better. Especially if the AutomountServiceAccountToken is available for both.

rogerio-kino-ifood:
@KnechtionsCoding done!

knmsk:
@KnechtionsCoding done!

KnechtionsCoding:
@knmsk If you want to come join us in: https://join.slack.com/t/k6io/shared_invite/zt-f243422e-mH0P3y9cpUCflwjSoLlG4Q we can chat with @simme and @dgzlopes about getting this merged. I just built another PR off of this one.

simme:
I think someone perhaps mentioned the wrong @simme ;) Just letting y'all know, since they might've missed the message!

KnechtionsCoding:
Whoops, sorry about that. Meant @simskij , my bad.

PR #51, opened 2021-07-20
Feat/yamllint github action

KnechtionsCoding:
Prevents bad yaml from being pushed

Includes lots of changes to bring yaml in compliance


dgzlopes:
Nice, thanks!

Issue #48, opened 2021-07-14
Allow for scheduled jobs


KnechtionsCoding:
Allow jobs to have a crontab schedule on them that causes a cronjob to be created instead of just a batch job.

This would allow for ongoing tests rather than just manually kicked off testing.

KnechtionsCoding:
I've put some time over the last two days into the cron job part. I think that this simply needs to be outside the operator.

I'll walk through why:

When trying to have the operator create k8s cron jobs directly.
* Marshalling and unmarshalling the yaml from the k6 object passed in becomes a pain. It's not directly transferrable to kubectl, so you either have strip fields, or hand map everything (if someone has a better way let me know).
* Permissions. The operator would have to declare a service account in each namespace that allows the cron job to create k6 objects. This involves a vast expansion of the permissions required, and not really tenable.

When trying to use the cron inside the operator, via go routines.

* Very difficult to solve for concurrency, and distributed controllers. Greatly increases complexity.
* Requires a lot of changes to how the code actually works in terms of scheduling inside. basically means, create new k6 thing on cron schedule, wait to reconcile, etc. Adds another loop to the process via the cron, not ideal.

I think the better answer is to create the cron jobs outside. I can update the documentation with an example for kustomize on how that works. This allows for individuals to control all aspects of their cronjob without diluting what the operator does.

Issue #22, opened 2021-06-04
Support for imagePullSecrets


KnechtionsCoding:
Ideally we can pass in an imagePullSecrets to the job to allow for images from custom docker repositories.

sandeep173254:
Hi @KnechtionsCoding  , can you please elaborate on how can we pass imagePullSecret to the job without modifying the code?

PR #52, opened 2021-07-21
Feat/service account

KnechtionsCoding:


KnechtionsCoding:
Provides documentation for #48 and provides support for serviceAccounts for #22 Image Pull Secrets are better managed that way.

pkg/resources/jobs/runner.go
```diff
@@ -23,7 +23,7 @@ type Script struct {
 // NewRunnerJob creates a new k6 job from a CRD
 func NewRunnerJob(k6 *v1alpha1.K6, index int) (*batchv1.Job, error) {
 	name := fmt.Sprintf("%s-%d", k6.Name, index)
-	command := []string{"scuttle", "k6", "run", "--quiet"}
```
dgzlopes, line 26: Do we want to remove the `--quiet` flag?

README.md
```diff
@@ -131,6 +174,119 @@ Note that we are replacing the test job image (`k6-prometheus:latest`), passing
 (`--out prometheus`), and also exposing the ports required for Prometheus to scrape the metrics
 (in this case, that's port `5656`)
 
+You'll also need to create a pod monitor:
```
dgzlopes, line 177: This only applies if the user is using the Prometheus Operator. Right?

I'm not very familiar with it...

dgzlopes (commented):
Thanks for the PR! Looking good :)

Left two small comments/questions.

README.md
KnechtionsCoding: Yes, I can update that.

pkg/resources/jobs/runner.go
KnechtionsCoding: Ah, good call out. I didn't mean to add that to this PR.

PR #54, opened 2021-07-22
Feat/paused quiet bool

KnechtionsCoding:
Build off of #49  and #53 this adds the option to start tests immediately, and allows for full logs to be outputted. Defaults stay the same. Passing tests on top of it.

PR #53, opened 2021-07-22
tests: write tests for jobs package (87.8%)

KnechtionsCoding:
Signed-off-by: Hans Knecht <Hans.Knecht@missionlane.com>

PR #56, opened 2021-07-26
Feat/optional-istio - Ignore the Istio process in the beginning

knmsk:
Added a way to ignore the scuttle inside starter and runner.

edit: Build off of #49, #53 and #54


KnechtionsCoding (commented):
Overall, I like this. Two general comments:

1. Should we put the ENV variables for scuttle behind a bool as well? This would prevent unnecessary ENV or edge case interactions. But might be a big pain.

2. Composition wise. Should we move the istio check into the helpers.go and let it return a command and env variables? So we can keep everything in the same place?

knmsk:
> Should we put the ENV variables for scuttle behind a bool as well? This would prevent unnecessary ENV or edge case interactions. But might be a big pain.
Probably it's better to create a struct for this stuff.

> Composition wise. Should we move the istio check into the helpers.go and let it return a command and env variables? So we can keep everything in the same place?
Sure, we can move into the helpers.go it makes sense!

KnechtionsCoding (approved):

PR #68, opened 2021-08-23
Added to readme section about cluster setup

MateuszGrab:
As I am trying to use k6, on k8s, I've found that there is lack of mentions, about Istio & Envoy being enabled on cluster. 
Here is fix to Readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ Read also the [complete tutorial](https://k6.io/blog/running-distributed-tests-o
 
 ## Setup
 
+### Cluster requirements
+To be able to spawn runner pods, k6-operator needs Istio enabled in namespace, where it operates.
+
 ### Deploying the operator
 Install the operator by running the command below:
 


KnechtionsCoding:
@MateuszGrab Istio isn't required. You can turn it off by not enabling scuttle.

MateuszGrab:
When using `make deploy` with defaults, it does complains in the logs about not being able to connect to to envoy. 
I had to install Istio, and enable that in a namespace to work. 
From what I saw it was having scuttle on pods as part of entrypoint as well.

KnechtionsCoding:
It should time out after the default of 15s and boot up anyway. Scuttle has a default timeout of 15s to check for envoy/istio prior to just running the default stuff. We could remove scuttle from the entrypoint, but the command is being set via the scuttle option in the spec anyway.

KnechtionsCoding:
@MateuszGrab I can't remember the default right now, but you can try setting scuttle.enabled: false on the k6 object and scuttle shouldn't be involved at all.

The command created by the operator as part of the pod spec for the runner and starter should override the entry point.

simskij:
Thanks for the PR @MateuszGrab, I really appreciate it! However, I don't think I want to merge this. If it indeed requires Istio to function, we need to change that rather than update the docs, as it's not intended (as you might have suspected based on @KnechtionsCoding's comments)

sgrzemski:
@simskij @KnechtionsCoding I don't think Istio should be enabled by default.

PR #76, opened 2021-09-29
feat: migrate from apiextensions.k8s.io/v1beta1 to apiextensions.k8s.…

KnechtionsCoding:
…io/v1

v1beta1 will no longer be supported in 1.22 and has been supported since 1.16


KnechtionsCoding:
I've validated that this doesn't change behavior, tests still run without modification, etc. But the important part would be to ensure that any PRs merged after this have the updated makefile and regenerate the CRD based on the v1 extension api

Issue #83, opened 2021-12-03
Update Golang to 1.17 along with dependencies


yorugac:
This involves Golang project and Dockerfile(s) both.

yorugac:
It's best to do this before #9 to have clean sequential merges.

PR #86, opened 2021-12-07
Support for the Cloud output

### LINKED ISSUE #9: Add support to ship results to the cloud
Here we're running a lot of k6 instances in parallel, but all of them are part of the same test.

I would like to ship the results from all these instances as a unique test run on the cloud.

yorugac:
This PR adds basic support for the Cloud output. If token is set and arguments contain `--out cloud`, the test will generate a new test run on the Cloud and send metrics to the Cloud. In other aspects, operator should behave as before.

Things of note:
- in case of the Cloud output there is some initialization and finishing tasks to do so new stages were added as well as initializing job which starts one pod with the same image as runner
- token is currently passed in as a k8s secret
- some refactoring was done here too, mostly to avoid code duplication, esp. around how script / configmap is handled
- the default labels for runner jobs now include `"runner": "true"` to be able to distinguish between them and initializing pod
- this implementation depends on [this PR to k6](https://github.com/grafana/k6/pull/2279) so it'll work only with k6 v0.36+

Current issues / things to improve:
- ~~thresholds are absent (a blocker on the cloud side)~~ was resolved

It'd be best to see to #83 before merging this PR.

Fixes #9


controllers/k6_initialize.go
```diff
@@ -0,0 +1,197 @@
...
+			return res, err
+		}
+
+		if t, ok := secrets.Items[0].Data["token"]; !ok {
+			err := fmt.Errorf("The secret doesn't have a field token for k6 Cloud")
+			log.Error(err, err.Error())
+			return res, err
+		} else {
+			token = string(t)
+		}
```
brian-groux-hs, line 71: Can we consider trying to load the token from other sources as well? For instance we typically don't use k8s secrets, instead we use Vault with a side car to mount a secrets volume for the container. In our case, with Vault, we were hoping to use the file configuration (https://k6.io/docs/cloud/integrations/token/#authentication-with-a-config-file). Is it possible to look for the other methods of providing a token as documented at the link above, then also look for k8s secret? While the env var approach is certainly not as secure, it does come in handy for testing...however, at the very least file config support would be appreciated. :)

yorugac: Hi @brian-groux-hs! Sorry for such a delay in response.
Yes, I definitely would like to have alternative options to load secrets here. I wasn't sure what people might use so you just gave me an answer to that, thank you :smile: I'm not sure at the moment if operator should try to load everything one by one as you describe or be a separate configuration option though...
Either way, I won't be able to add it right now but noted in TODO. I'll probably create a separate issue for this.

yorugac: Related issue: https://github.com/grafana/k6-operator/issues/103

Issue #87, opened 2021-12-08
Switch from REST API to k6 CLI in starter job


yorugac:
The starter is currently relying on REST API of k6 to start the runners. It's workable but the more "officially supported" way is to use the CLI, e.g. `k6 resume`. Investigate and switch to using the CLI for the starter.

Initiating pod in https://github.com/grafana/k6-operator/pull/86 can be used as an additional baseline here and/or for refactor.

simskij:
Just thought I'd offer some background as we (read: I) opted for using the HTTP Rest API over going for direct pod invocation deliberately.

While using `k6 resume` would work, it would require higher privileges for the starter pod, which I'm not really sure is an excellent idea. Allowing the starter pod to execute HTTP requests is by far less likely to cause issues than giving a pod the privileges to execute arbitrary shell commands in the load generators.

yorugac:
Hi @simskij, thank you for providing additional background and input :slightly_smiling_face: 

I don't think I'm following here: what do you mean by higher privileges? Internally, `k6 resume` is nothing but a wrapper around REST call.

simskij:
My bad! I was interpreting your intention as calling `k6 resume` **inside** the runners using the programmatic equivalent to `kubectl exec`, which would then have required the starter to have execution privileges in the runner pods.

yorugac:
This might not be fully viable after all :disappointed: given cases like https://github.com/grafana/k6-operator/pull/89

frittentheke:
@yorugac  @simskij 

While I totally understand and appreciate the synchronization to start all k6 instances (`parallel`>1) I am wondering why the operator does not simply issue those HTTP calls (
https://github.com/grafana/k6-operator/blob/40dce3ffc13c55fb665c41b891c1459d3ad66068/pkg/resources/containers/curl_start.go#L29) to the REST API endpoints natively? It just seems rather complex to spin up another job / pod to then just issue a few HTTP calls via curl.

Is this is to work around potential network policies going when going across namespaces or what is the intention here?

yorugac:
> I am wondering why the operator does not simply issue those HTTP calls to the REST API endpoints natively? It just seems rather complex to spin up another job / pod to then just issue a few HTTP calls via curl.

Given the discussion above, I believe this is what this issue is basically about at this point :smile: I should update the description to reflect that.

This issue wasn't prioritized so far simply because nobody seemed to care, until recently. So thank you greatly for sharing your opinion @frittentheke! 

[...]

PR #89, opened 2022-01-06
fix: improve test startup logic

victorlcm:
Hello everyone,

We've been using this operator for a while and we observed a strange behavior on tests that took a long time on init phase. We noticed that they remained in "paused state" because the k6 http server was not ready yet. So, we've added a simple verification to make sure the http server is up before sending the start requests to the jobs. It's a simple fix, but the results were very effective.

### DIFF
--- a/controllers/k6_start.go
+++ b/controllers/k6_start.go
@@ -3,6 +3,8 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -14,6 +16,22 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+func isPodReady(pod *v1.Pod) bool {
+	resp, err := http.Get(fmt.Sprintf("%v/v1/status", pod.Status.HostIP))
+
+	if err != nil {
+		return false
+	}
+
+	status, err := strconv.Atoi(resp.Status)
+
+	if err != nil {
+		return false
+	}
+
+	return status < 400
+}
 
 // StartJobs in the Ready phase using a curl container
 func StartJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.K6, r *K6Reconciler) (ctrl.Result, error) {
@@ -35,7 +53,7 @@ func StartJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.K6, r *K6Recon
 
 		var count int
 		for _, pod := range pl.Items {
-			if pod.Status.Phase != "Running" {
+			if pod.Status.Phase != "Running" && !isPodReady(&pod) {
 				continue
 			}
 			count++
--- a/controllers/k6_start.go
+++ b/controllers/k6_start.go
@@ -17,6 +17,9 @@ import (
 )
 
 func isPodReady(pod *v1.Pod) bool {
+        if len(pod.Status.HostIP) < 1 {
+            return false
+        }
 	resp, err := http.Get(fmt.Sprintf("%v/v1/status", pod.Status.HostIP))
 
 	if err != nil {
--- a/controllers/k6_start.go
+++ b/controllers/k6_start.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"strconv"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -24,13 +23,7 @@ func isPodReady(pod *v1.Pod) bool {
 		return false
 	}
 
-	status, err := strconv.Atoi(resp.Status)
-
-	if err != nil {
-		return false
-	}
-
-	return status < 400
+	return resp.StatusCode < 400
 }
 
 // StartJobs in the Ready phase using a curl container
--- a/controllers/k6_start.go
+++ b/controllers/k6_start.go
@@ -16,13 +16,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func isPodReady(pod *v1.Pod) bool {
-        if len(pod.Status.HostIP) < 1 {
-            return false
-        }
-	resp, err := http.Get(fmt.Sprintf("%v/v1/status", pod.Status.HostIP))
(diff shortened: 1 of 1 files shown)


yorugac:
Hi @victorlcm, thank you for catching this case and fixing it! :clap: 

Just as a heads-up, I'll be able to fully test the logic of this PR only on the next week. But at first glance, it looks fine to me :+1:

controllers/k6_start.go
```diff
@@ -15,6 +17,22 @@ import (
...
 )
 
+func isPodReady(pod *v1.Pod) bool {
+	resp, err := http.Get(fmt.Sprintf("%v/v1/status", pod.Status.HostIP))
+
+	if err != nil {
+		return false
+	}
+
+	status, err := strconv.Atoi(resp.Status)
```
na--, line 27: You can just use `resp.StatusCode`, which is already `int`

victorlcm: Thanks! Just fixed it

controllers/k6_start.go
```diff
@@ -15,6 +16,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+func isPodReady(pod *v1.Pod) bool {
+	resp, err := http.Get(fmt.Sprintf("%v/v1/status", pod.Status.HostIP))
```
yorugac, line 20: It makes sense to add this check here:
```suggestion
        if len(pod.Status.HostIP) < 1 {
            return false
        }
	resp, err := http.Get(fmt.Sprintf("%v/v1/status", pod.Status.HostIP))
```
At first, this field is empty so there is no point in making any network requests then.

yorugac (changes requested):
@victorlcm I've been looking at this PR more closely today and I wish to clarify with you the following: in your use case, do you run it with `separate` flag in yaml configuration?

So far, all seems good, though there is one more thing I request you to adjust, please see the comment :slightly_smiling_face: I'll check a few more things in the meantime.

victorlcm:
Hey @yorugac, we're running with separate `false`, while controlling how much resources each job uses by  setting the runner limits and requests and limiting the VUs according to these resources 😄

yorugac:
Hi @victorlcm, thanks for the fix but I'm afraid that's not all yet :joy: I've found one more issue in current approach and we really should try to fix the logic there. Let me explain. Operator [uses k6's REST API](https://github.com/grafana/k6-operator/blob/d268e0e371d95f51f3ee0ca73da61daa3d206544/pkg/resources/containers/curl.go#L27-L29) for starting runners but it does it with hostnames, not IPs. Hostnames come from:
https://github.com/grafana/k6-operator/blob/d268e0e371d95f51f3ee0ca73da61daa3d206544/controllers/k6_start.go#L59-L61

In contrast, your PR relies on IP of the runner. This approach is incorrect for the cases when there is more than 1 runner on the same node. I assume you haven't encountered such an error during your usage but it can happen otherwise. Basically the check for `isPodReady` should become `isServiceReady` and it can be executed right in the above loop. I.e. if at least 1 service is not responding, then log error and `return false`.

Then the checks become separate:
1) get all pods and see if they are in `Running` state
2) get all services and see if they respond to REST
3) now start

Does it makes sense? Could you please try this updated logic with your use case, to see if it continues to work?
I'm sorry for causing confusion but I've spotted this issue only during testing with your branch recently :sweat_smile:

victorlcm:
> Hi @victorlcm, thanks for the fix but I'm afraid that's not all yet 😂 I've found one more issue in current approach and we really should try to fix the logic there. Let me explain. Operator [uses k6's REST API](https://github.com/grafana/k6-operator/blob/d268e0e371d95f51f3ee0ca73da61daa3d206544/pkg/resources/containers/curl.go#L27-L29) for starting runners but it does it with hostnames, not IPs. Hostnames come from:
> 
> https://github.com/grafana/k6-operator/blob/d268e0e371d95f51f3ee0ca73da61daa3d206544/controllers/k6_start.go#L59-L61
> 
> In contrast, your PR relies on IP of the runner. This approach is incorrect for the cases when there is more than 1 runner on the same node. I assume you haven't encountered such an error during your usage but it can happen otherwise. Basically the check for `isPodReady` should become `isServiceReady` and it can be executed right in the above loop. I.e. if at least 1 service is not responding, then log error and `return false`.
> 
> Then the checks become separate:
> 
> 1. get all pods and see if they are in `Running` state
> 2. get all services and see if they respond to REST
> 3. now start
> 
> Does it makes sense? Could you please try this updated logic with your use case, to see if it continues to work? I'm sorry for causing confusion but I've spotted this issue only during testing with your branch recently 😅

Hey @yorugac, I'm sorry for the delay, this is a busy week for me. Your point makes sense! I'll make the changes later this week or early next week, is that ok?

Thanks for the great review!

yorugac:
@victorlcm NP, that'll be great! Looking forward to your update :slightly_smiling_face:

victorlcm:
hey @yorugac, I've changed to use the service approach, it seems to be working fine now, can you test it again? Tks! 😄

knmsk:
@yorugac the way this PR is, if this get approved, with more changes this can also remove the starter pod and add the logic on the operator to make a curl on the runners?

yorugac:
Hi @knmsk, good question. I think what you described is an option to move away from Curl completely (#87) but it'd result in removing a sync point from logic. I.e. right now, starter uses basically one command and acts as a synchronization point and removing it would create potential delays in when runners actually start. AFAIK, some setups might depend on starter having a separate security requirements (e.g. see #74). These points might be negligible in most cases or they might depend on exact implementation but I believe implications should be thought through before going that route.
Do you have a specific case in mind for that? Perhaps, it might make sense to create a separate issue to open a discussion on topic :slightly_smiling_face:

knmsk:
I will create a separated issue for this topic :)

yorugac (approved):
Hi @victorlcm, thank you for the update! I've tested it yesterday and I think it's good to go now :+1: :smile: 

There is one thing about domains that I wonder about: starters right now use a short form with `http://${service.ObjectMeta.Name}:6565` :thinking: But in this PR's check namespace is likely needed indeed... ATM, I'm not aware if this can fail anywhere but we should probably make a consistent scheme for DNS sometime.

Thanks again for all your work!

PR #91, opened 2022-01-14
[feat] Add option for disabling fragmentation

### LINKED ISSUE #90: [Feature Request] Allow disabling of execution segments when parallelism > 1
Great operator, thanks for making it!

I would like some fine-grain control on the runners that are created. I noticed that when parallelism is > 1, the number of VUs is shared across all the workers. My number of VUs is somewhat dynamic, and I currently have runners quitting prematurely because there is no work to do. 

It would be great if I could prevent this from happening.

jdheyburn:
Happy to rename the option to something more suitable. Status quo should be the default (fragment if parallelism > 1). User can disable by setting `disableFragmentation: true` in their spec.

Closes #90

### DIFF
--- a/api/v1alpha1/k6_types.go
+++ b/api/v1alpha1/k6_types.go
@@ -50,16 +50,17 @@ type K6Scuttle struct {
 
 // K6Spec defines the desired state of K6
 type K6Spec struct {
-	Script      K6Script               `json:"script"`
-	Parallelism int32                  `json:"parallelism"`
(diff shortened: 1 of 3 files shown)


yorugac:
Closing this PR: it's not going to be merged and it's been awhile :sweat_smile: The related [issue](https://github.com/grafana/k6-operator/issues/90) remains open: if someone would like to argue for adding a similar logic to k6-operator, please comment and vote in the issue!

Issue #75, opened 2021-09-27
Avoid test be executed again when it fails.


b0nete:
Hi, i'm executing load tests in my kubernetes cluster but i have a problem when tests fails.

I need tests be executed only one time, and if these run succesfully o fails don't be executed again. 
Currently if tests running ok these dont be executed again, but if test threshold faild automatically starter container is created and launch another pod to try run test again.

I leave my config files here, i tried to set abortOnFail in threshold and use abortTest() function but the problem persist.
I think it is a k6-operator behaviour, maybe you can help me.

This is my test file.
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: k6-test
  namespace: k6-operator-system
data:
  test.js: |
    import http from 'k6/http';
    import { Rate } from 'k6/metrics';
    import { check, sleep, abortTest } from 'k6';

    const failRate = new Rate('failed_requests');

    export let options = {
      stages: [
        { target: 1, duration: '1s' },
        { target: 0, duration: '1s' },
      ],
      thresholds: {
        failed_requests: [{threshold: 'rate<=0', abortOnFail: true}],
        http_req_duration: [{threshold: 'p(95)<1', abortOnFail: true}],
      },
    };

    export default function () {
      const result = http.get('http://test/login/');
      check(result, {
        'http response status code is 200': result.status === 500,
      });
      failRate.add(result.status !== 200);
      sleep(1);
      abortTest();
    }
```

And this is my k6 definition.
```
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6-sample
  namespace: k6-operator-system
spec:
  parallelism: 1
  script:
    configMap:
      name: k6-test
      file: test.js
  arguments: --out influxdb=http://influxdb.influxdb:8086/test
  scuttle:
    enabled: "false"
```

I hope you can help me, thanks!

KnechtionsCoding:
So, I think this is because k6 exits with a non 0 exit code and so the k6 operator will try to keep it going till it succeeds.

We could probably add that to the crd as an option, restart never. And have k6-operator interpret that.

yorugac:
@b0nete thanks for opening the issue!

[...]

PR #73, opened 2021-09-04
Feat/service mesh options

KnechtionsCoding:
Building off of #71 to allow for broader serviceMesh support

Closes #58

PR #93, opened 2022-01-17
Feat: Execution of test using a LocalFile inside the Docker Image

knmsk:
Hello everyone,

Recently i've had some problems running with a configMap due to the size of my test, the company that i work on doesn't like the idea of using VolumeClaim for applications in our clusters, so i made a simple fix to execute a local file inside the docker image.


config/samples/k6_v1alpha1_k6_with_localfile.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: k6.io/v1alpha1
+kind: K6
+metadata:
+  name: k6-sample
+spec:
+  parallelism: 4
+  script:
+    localFile: /test/test.js
```
yorugac, line 9: Please correct me if I'm wrong but this example makes sense when there is also a runner image defined right? So for completeness, the example should also have this part:
```
  runner:
      image: <custom-image>
```

yorugac (changes requested):
Hi @knmsk, thank you for this PR!
I didn't imagine such company restrictions before and even curious how often it happens elsewhere :smile: 

I've tested the changes locally and all seems to work fine; I have a small request to update an example, please see my comment.

I also wonder about one potential improvement of the logic here: is it possible to somehow check if `LocalFile` exists before launching runners? Right now they'll be launched and status will be unknown until starter's failure. I suppose one way is to launch one "pseudo-runner" first and check the files there but that seems quite unwieldy. Perhaps there is a better way? :thinking: I'm not sure about solution here at the moment so this is not a "change request" but in case you have ideas, please share :slightly_smiling_face:

knmsk:
Hi @yorugac, i've added the custom-image on the samples.
About the improvement of logic, i was thinking about in the last few hours, maybe we can add a command of `cat /path/test.js` to check if the file exists in case of failure probably the runner pod will be marked as Error and the starter will not run. But at the same time it should be better to notify the operator-controller-manager to mark them as error/completed. But this also involves the state of the CRD k6. What do you think?

yorugac:
@knmsk thanks for the update!

Re improvement. :thinking: so basically, launch the runner pods but if there's no script, they'll error out quickly? I think this may be a nice workaround... Perhaps, not "ideal" as we are still launching runners but better than nothing for sure :smile: Thank you very much for putting your time into this!
About operator notifications: we actually have a bug on that, here #75 Once we fix that bug, operator will likely have additional info about runners state and that should allow to add the relevant part here more easily. So at the moment, it seems to me that its best to postpone adding state changes until then.

Given that, to sum up: here, it's enough to only modify the command in case of `LocalFile` as you suggest for now. Runner will be restarting probably but that's a problem for another PR. How does that sound to you?

knmsk:
@yorugac sounds good to me!
> Re improvement. thinking so basically, launch the runner pods but if there's no script, they'll error out quickly?

Yeah that's the idea.

yorugac (approved):
LGTM!

yorugac:
Hi @knmsk, just for my understanding; do you plan to add improvement to the runner's command in case for `LocalFile`s? :slightly_smiling_face:

knmsk:
@yorugac hi, i've been trying to implements the file verification. Found a way but it's kind a workaround. We possibly can change this to always have some kind of .sh file to not let this kind of messy command. The runner command would be changed:
**FROM**:
```yaml
command:
  - k6
  - run
  - /test/test.js
```

**TO**:

```yaml
command:
  - sh
  - -c
  - >-
     if [ ! -f "/test/taest.js" ]; then echo "Local file doesn't exist"; exit 1; fi;
     k6 run "/test/test.js"
```
What's your opinion about this? 
This will only applies when you use the LocalFile options at least what i have been thinking.

yorugac:
@knmsk thanks for response! yes, it's a bit messy :disappointed: But TBH, I don't really think it can be done otherwise... Since runner image is custom, there is no guarantee that `.sh` file will be present. Unless we come up with some way to scan runner image for files before launching anything in operator? But adding such a thing at this point would be even more messy and even if we'll add such a scan for other reasons in the future, it is definitely not happening now. So yeah, I don't think we can avoid these complications :sweat_smile: 

So I think let's add this check now. I'm going to add some refactoring around script & volumes (partially already done in #86 and an "inevitable" addition with or after Hans' work on #73). And it _should_ be possible to incorporate this check in that refactoring too, to have a more readable and easily changeable code around the additions. So the plan then is: add a check in this PR, merge it in, and extend "refactoring plan" currently in #86 to commands as well.

As always, if there are additional thoughts and ideas, I'll be happy to hear them :slightly_smiling_face:

knmsk:
@yorugac done!

yorugac:
Thank you @knmsk ! LGTM :+1:

Issue #123, opened 2022-06-30
Feature request: can be mount ConfigMap/Secret to k6 pods


mugioka:
# motivation
I am dealing with binaries (jpeg images) in k6 script as shown below. However, I am having trouble mounting the volume to k6 pods with my current K6 CR.
```js
import http from 'k6/http';
import { check, group, sleep } from 'k6';

export const options = {
  vus: 10,
  duration: '300s',
  iterations: 100000,
};

const SLEEP_DURATION = 0.1;

const miniImgFile = open('./test-mini.jpg', 'b');

const bigImgFile = open('./test-big.jpg', 'b');

export default function () {
  // omit
}
```

yorugac:
Hi @mugioka,
Sorry for the delay. Thank you for both the issue and the PR! Could you please explain why `volumeClaim` option doesn't work for you? (I've just added a section to the README for it, since it wasn't very explicit before: [here](https://github.com/grafana/k6-operator/blob/main/README.md#volumeclaim))

mugioka:
Thank you for reply @yorugac 
If I understand correctly, it is my understanding that the VC specified in the README can only be used to mount k6 scenario.
If I want to mount images base64 encoded and mounted in a container, I need to change this PR.

yorugac:
> If I understand correctly, it is my understanding that the VC specified in the README can only be used to mount k6 scenario.

No, it doesn't have such limitations, it's just a normal volume. More precisely, runner does expect to find the script there but you can store any other data as well, e.g. any additional files like JSONs or images, etc. If the script uses those files, then k6 would reference the mounted folder the same way it does normally. Example:
```
# volume contains:
test.js
users.json
```
Then your `test.js` can load that JSON as `JSON.parse(open("./users.json"))`

yorugac:
@mugioka I see, I was looking at your PR and was thinking about volumes themselves mostly. But if the problem is with secrets, then `secretKeyRef` should work even now:
https://github.com/grafana/k6-operator/blob/c63b127b53ca18ddf98f1648e25aefb785efe207/config/crd/bases/k6.io_k6s.yaml#L791
So it should be possible to access `Secret` as an env var. It looks like it is present in `envFrom` as well, by the way.

Also, if the key is in `ServiceAccount`, it can be passed directly as described here:
https://github.com/grafana/k6-operator/blob/main/README.md#serviceaccount

Those are the options now, I believe.

mugioka:
@yorugac 
[...]

PR #124, opened 2022-06-30
feat: can be mount the volumes to runner/starter pods.

mugioka:
Closes https://github.com/grafana/k6-operator/issues/123.


config/crd/bases/k6.io_k6s.yaml
```diff
@@ -1944,6 +1944,725 @@ spec:
...
+                          type: string
+                        subPathExpr:
+                          type: string
+                      required:
+                      - mountPath
+                      - name
+                      type: object
+                    type: array
+                    x-kubernetes-list-type: atomic
+                  volumes:
```
mugioka, line 1968: https://github.com/kubernetes/kubernetes/blob/master/api/openapi-spec/v3/apis__apps__v1_openapi.json#L4413

config/crd/bases/k6.io_k6s.yaml
```diff
@@ -1944,6 +1944,725 @@ spec:
                     type: object
                   serviceAccountName:
                     type: string
+                  volumeMounts:
```
mugioka, line 1947: https://github.com/kubernetes/kubernetes/blob/master/api/openapi-spec/v3/apis__apps__v1_openapi.json#L4679

Issue #122, opened 2022-06-30
Feature request: can be set environment variable from envFrom in k6 CR.


mugioka:
# motivation
When executing k6 scripts, we need to handle sensitive information such as API Key in environment variables, but currently we cannot define environment variables in envFrom, so we are having trouble handling sensitive information.
By enabling envFrom to be used with k6 CR, secret information can be managed with secret resources, making it possible to handle secret information more securely.

BarthV:
https://github.com/grafana/k6-operator/blob/main/api/v1alpha1/k6_types.go#L53-L65

K6 specs `Runner` key is a standard kubernetes pod spec, So you can perfectly add standard pod env specs usign any existing configMap or Secret.

An example from my own test suites :
```
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6
spec:
  parallelism: 2
  script:
    localFile: /k6/pipeline.js
  arguments: --stage 120s:180 --stage 300s:360 --stage 420s:360 --stage 60s:0
  runner:
    env:
    - name: FOOBAR
      value: http://foor.far
    - name: SECRET
      valueFrom:
        secretKeyRef:
          name: my-api-secret
          key: secret
    - name: SOME_VAR
      value: any-value
```

mugioka:
Ah, sure, but I think it would be better to be able to use envFrom if you want to handle multiple environment variables.

yorugac:
I'm inclined to agree that `env` is sufficient in many cases, esp. since k6-operator is fully on-premise deployment anyway. So it seems like `envFrom` is more about convenience than anything else. But that's important too and I'll be happy to merge it as a PR. Thank you :smile:

PR #129, opened 2022-07-15
feat: ability to set the environment variables with envFrom in k6 runner/starter.

mugioka:
Closes https://github.com/grafana/k6-operator/issues/122.

### DIFF
--- a/config/crd/bases/k6.io_k6s.yaml
+++ b/config/crd/bases/k6.io_k6s.yaml
@@ -988,13 +988,35 @@ spec:
                     type: object
                   serviceAccountName:
                     type: string
+                  envFrom:
+                    type: array
+                    items:
+                      description: K6EnvFrom represents the source of a set of ConfigMaps/Secrets
+                      default: {}
+                      properties:
+                        configMapRef:
+                          type: object
+                          description: The ConfigMap to select from
+                          properties:
+                            name:
+                              type: string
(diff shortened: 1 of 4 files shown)


config/crd/bases/k6.io_k6s.yaml
```diff
@@ -988,13 +988,35 @@ spec:
                     type: object
                   serviceAccountName:
                     type: string
+                  envFrom:
```
mugioka, line 991: definition is [here](https://github.com/kubernetes/kubernetes/blob/7e2c02d4c86d73051b6d4f47298d35a7ffb42b6a/api/openapi-spec/v3/apis__apps__v1_openapi.json#L1681)

yorugac: Could you please add this as a comment?

yorugac (changes requested):
@mugioka Thanks for the PR! It LGTM :+1: I'll run a couple of tests but should be good to merge. Please see there is a small request change.

config/crd/bases/k6.io_k6s.yaml
mugioka: I fixed it with https://github.com/grafana/k6-operator/pull/129/commits/1559ba0f183fbc9c89fadd736f1dd17dc78bd354.

PR #135, opened 2022-08-02
feat: use name instead of index for instance id tag

spoukke:
Currently,  the tag used for instance id is the index of the job. But, if you run multiple K6 resources at the same time, there is no way to distinguish them.

Here I propose to use the name of the job, instead of the index of the job


yorugac:
Hi @spoukke,
Thanks for the PR and sorry for the delay! We cannot change `instance_id`: the whole point of `instance_id` tag is to distinguish between pods, not jobs. I suggest changing it to `job_name` tag instead and leaving `instance_id` as is. Would it work for your case?

spoukke:
Yes, it would work, I am going to do the modification!

yorugac (approved):
LGTM :+1: Thank you @spoukke!

yorugac:
But not the unit tests, apparently: https://github.com/grafana/k6-operator/runs/7917488949?check_suite_focus=true
They check the command and new tag is part of the command. Please check.

spoukke:
Hi @yorugac ! Just updated the unit tests

yorugac (approved):
@spoukke Thanks for following up with this! Merging in now.

Issue #132, opened 2022-07-27
k6-controller starter doesn't find service pods


pekavau:
Hi!

On our developer machines, we have the issue that when we start the operator, the starter pod tries to make http GET calls to the k6 services but can't dns resolve them.

We've tried this on 4 separate machines, with a mix of linux/windows and docker for desktop / minikube and get the same results.

When we manually do an nslookup from the starter pod, it times out, and ping doesn't find the services.
Manually adding the domain "<namespace>.svc.cluster.local" _does_ enable nslookup to find the services.

It _does_ work after we disconnect the local machine from the network (!)

We don't know if it's related, but the IP addresses of the service nodes and the other pods could be in conflict with the ip addresses that our internal network devices are using (all in the 10.*.*.* range)

I'm attaching the starter logs, the dns logs, a successful nslookup, and the list of services. Let me know if you need more details, Cheers, pekavau.

[coredns-6d4b75cb6d-62mf8.log](https://github.com/grafana/k6-operator/files/9199423/coredns-6d4b75cb6d-62mf8.log)

[k6-sample-starter-94jcd.log](https://github.com/grafana/k6-operator/files/9199457/k6-sample-starter-94jcd.log)

![Screenshot 2022-07-27 154007](https://user-images.githubusercontent.com/32483753/181262456-93fdb903-f956-42be-ad8a-84eb540bb101.png)

![Screenshot 2022-07-27 153606](https://user-images.githubusercontent.com/32483753/181262486-569c53a2-93ff-4f83-92dd-052e523130ad.png)

pekavau:
Note:
The attached diff works around the issue

Also, on my linux machine, when I switch minikube from the kvm2 virtualization driver to the docker driver, it also works (doesn't work with kvm2 driver)

yorugac:
@pekavau IIRC, something similar came up on the community forum once before but it wasn't resolved then. Thank you for the issue and finding the fix! Would you like to make a PR?

PR #140, opened 2022-08-10
Fix: use Cluster IP instead of service name

danielpalstra:
Uses the Cluster IP to connect to services instead of the service name. Credits to @pekavau for finding the issue and workaround.

fixes  #132


yorugac (approved):
Thanks @danielpalstra for putting this fix up :slightly_smiling_face:

PR #142, opened 2022-08-10
feat: add use of tolerations for starter and runner

spoukke:
This PR aims to add support for Tolerations for the starter and the different runner pods.


yorugac (changes requested):
Hi @spoukke, thanks for the PR! Could you please update manifests as well? We have `make manifests` to help with that.

yorugac:
And one more thing: I think tolerations should be added to all pods in the operator, i.e. to the initializer pod as well.

spoukke:
Hi @yorugac ! I re-generated the manifests.

From what I see in the codebase, the Initializer is of type `*batchv1.Job` so it will directly inherit the filed Tolerations from `corev1.PodTemplate`.

Also, I was wondering why you did not use the `corev1.PoTemplate` for the Runner et Starter fields instead of recreating your own struct. If you do this, you will directly have all the fields form a Pods, and the operator will automatically follow the updates from the Kubernetes API.

spoukke:
Hello @yorugac ! I juste understood what you wnted regarding the initializer. Now, the runner tolerations are mapped to the intializer one.

Let me know if there is anything else

PR #151, opened 2022-08-28
Updates README and adds an example for volumeClaim testing usage

ionicc:
## What does this PR do?
* Adds a volumeClaim example: `config/samples/k6_v1alpha1_k6_with_volumeClaim.yaml`.
* Updates the readme with a more thorough explanation on how to use volumeClaims and also adds a directory example.

## Why?
* After taking some time debugging and experimenting with running k6 tests using volumeClaims, I believe having more concrete examples and explanations like this would help folks working with it in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,18 +41,58 @@ $ kubectl create configmap my-test --from-file /path/to/my/test.js
 
 #### VolumeClaim
 
-If you have a PVC with the name `stress-test` containing your script and any other supporting file(s), you can pass it to the test like this:
+There is a sample avaiable in `config/samples/k6_v1alpha1_k6_with_volumeClaim.yaml` on how to configure to run a test script with a volumeClaim.
+
+If you have a PVC with the name `stress-test-volumeClaim` containing your script and any other supporting file(s), you can pass it to the test like this:
 
 ```
 spec:
   parallelism: 2
   script:
     volumeClaim:
-      name: "stress-test"
+      name: "stress-test-volumeClaim"
       file: "test.js"
 ```
 
 ***Note:*** the pods will expect to find script files in `/test/` folder. If `volumeClaim` fails, it's the first place to check: the latest initializer pod does not generate any logs and when it can't find the file, it will terminate with error. So missing file may not be that obvious and it makes sense to check it manually. See #143 for potential improvements.
+
+##### Example directory structure while using volumeClaim
+
+```
+├── test
+│   ├── requests
+│   │   ├── stress-test.js
+│   ├── test.js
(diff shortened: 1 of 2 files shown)


config/samples/k6_v1alpha1_k6_with_volumeClaim.yaml
```diff
@@ -0,0 +1,16 @@
...
+  parallelism: 4
+  script:
+    volumeClaim:
+      name: stress-test-volumeClaim
+      # test.js should exist inside /test/ folder.
+      # And, All the js files and directories test.js
+      # is importing from should be inside the same directory as well.
+      file: test.js
+  runner:
+    image: <custom-image>
```
yorugac, line 16: Custom image is not a requirement for `volumeClaim`: could you please remove it from the example? So that there's no confusion about it.

README.md
```diff
@@ -41,19 +41,59 @@ $ kubectl create configmap my-test --from-file /path/to/my/test.js
...
 
+##### Example directory structure while using volumeClaim
+
+```
+├── test
+│   ├── requests
+│   │   ├── stress-test.js
+│   ├── test.js
+```
+In the above example, `test.js` imports a function from `stress-test.js` and, They would look like this:
```
yorugac, line 67:
```suggestion
In the above example, `test.js` imports a function from `stress-test.js` and they would look like this:
```

yorugac (changes requested):
@ionicc Thank you a lot for this PR! No doubt that it'll definitely be helpful :+1: :smile: 

I have a bit of change request; could you please update it?

config/samples/k6_v1alpha1_k6_with_volumeClaim.yaml
ionicc: @yorugac Thanks for noticing that, Removed it now!

README.md
ionicc: @yorugac Nicely observed, Fixed!

yorugac (approved):
@ionicc Great, thanks for the update!

patenvy1234:
how to create a volume claim pls explain

yorugac:
@patenvy1234, creating volumes and volume claims is part of standard Kubernetes docs:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/#create-a-persistentvolume
There're also a lot of third-party guides and howto's in the network about it.

patenvy1234:
I have already created volume claim now how store the test.js inside it ? and what should be the mounting location

PR #166, opened 2022-11-15
Do not add annotations to initializer based upon runner specification

### LINKED ISSUE #164: The initializer can't complete when there is more than one container running within the same pod
## Symptom

When running more than one container together within the same pod as the initializer, the operator can't retrieve container logs because the [container was not specified](https://github.com/grafana/k6-operator/blob/a7c4a3f2ea7341f2ffaa9c28af70f82ebaf53c84/controllers/k6_initialize.go#L110).

## Explained

I'm trying to use a custom k6 runner pod with [dapr](http://github.com/dapr/dapr) configured, in order to get this running, I have to add the dapr annotations to the runner, as the following:

```yaml
dapr.io/app-id: "tester-app"
dapr.io/enabled: "true"
```

As the initializer pod annotations are the [same as specified for runners](https://github.com/grafana/k6-operator/blob/main/pkg/resources/jobs/initializer.go#L35), the initializer pod will also contain the daprd container, leading in two different problems:

1) As k8s jobs won't complete until all containers are done, the dapr will make this fail because I have no control to shutdown the pod, and, [as the per documentation](https://docs.dapr.io/operations/hosting/kubernetes/kubernetes-job/), the correct way for doing this is calling sidecar shutdown.

A simple solution would be having different annotations for initializer and runners because the runner itself can do http requests on teardown but the initializer doesn't.

2) Even though a patched k6 version is used for shutting down the dapr sidecar whenever the initialization is complete, the operator fails as it is unable to read the pod logs.

## Suggested solutions

1) Do not reuse the same annotations as used for the test runner
2) Specify the k6 container when reading pod logs

latest

### K6 YAML

```yaml
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6-sample
spec:
  parallelism: 1
  script:
    configMap:
      name: tests
      file: test.js
  runner:
    metadata:
      annotations:
        dapr.io/app-id: "tester-app"
        dapr.io/enabled: "true"
    image: localhost:5001/k6-custom:latest
```

dapr installed

```sh
git clone github.com/mcandeia/k6-operator-dapr-error
cd k6-operator-dapr-error
./setup.sh
```

Check the manager logs:

```
2022-11-05T00:47:43.891Z        ERROR   controllers.K6  unable to stream logs from the pod      {"k6": "default/k6-sample", "error": "a container name must be specified for pod k6-sample-initializer-cqs8s, choose one of: [k6 daprd]"}
```

### Expected behaviour

initializer should complete regardless of running containers

### Actual behaviour

The test initializer can't complete

https://github.com/grafana/k6-operator/issues/164#tasklist-block-8de0afeb-65ae-40a0-87b0-bcd26f483ef3

javaducky:
Thanks to the awesome investigation/write-up by @mcandeia, this update no longer copies annotations from the runner specification. Each runner pod will still receive annotations as desired. If annotations support in initializer pod is desired, we'll probably need to incorporate an `Initializer Pod` entry within the [`K6Spec`](https://github.com/grafana/k6-operator/blob/main/api/v1alpha1/k6_types.go#L63-L64).

Signed-off-by: javaducky <javaducky@gmail.com>

Fixes #164

### DIFF
--- a/pkg/resources/jobs/initializer.go
+++ b/pkg/resources/jobs/initializer.go
@@ -20,7 +20,6 @@ func NewInitializerJob(k6 *v1alpha1.K6, argLine string) (*batchv1.Job, error) {
 
 	var (
 		image                        = "ghcr.io/grafana/operator:latest-runner"
-		annotations                  = make(map[string]string)
 		labels                       = newLabels(k6.Name)
 		serviceAccountName           = "default"
 		automountServiceAccountToken = true
@@ -29,10 +28,6 @@ func NewInitializerJob(k6 *v1alpha1.K6, argLine string) (*batchv1.Job, error) {
 
 	if k6.Spec.Runner.Image != "" {
 		image = k6.Spec.Runner.Image
-	}
-
-	if k6.Spec.Runner.Metadata.Annotations != nil {
-		annotations = k6.Spec.Runner.Metadata.Annotations
 	}
 
 	if k6.Spec.Runner.Metadata.Labels != nil {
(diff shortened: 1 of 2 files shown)


mcandeia (approved):

mcandeia:
Awesome @javaducky thanks for working on it!

PR #170, opened 2022-11-23
Add runner=true label to matchedExpressions in podAntiAffinity

### LINKED ISSUE #168: k6.spec.separate also applies affinity to starter-pod
### Brief summary

Hi, 

it seems that when separating runners to different nodes, the starter-pod is also been scheduled on its own node. Is this intended?

0.0.8

### K6 YAML

apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6-test
spec:
  parallelism: 2
  separate: true

_No response_

Deploy k6 CRD with separate=true

### Expected behaviour

AntiAffinity should only affect runner-pods not initializer and starter

### Actual behaviour

AntiAffinity is applied to runner and starter pods

cmergenthaler:
Hi,
this PR fixes scheduling of starter-Pod when using `spec.separater=true` by adding `runner=true` to matched Label-Expressions in podAntiAffinity.

Fixes #168

### DIFF
--- a/pkg/resources/jobs/runner.go
+++ b/pkg/resources/jobs/runner.go
@@ -231,6 +231,13 @@ func newAntiAffinity() *corev1.Affinity {
 									"k6",
 								},
 							},
+							{
+								Key:      "runner",
+								Operator: "In",
+								Values: []string{
+									"true",
+								},
+							},
 						},
 					},
 					TopologyKey: "kubernetes.io/hostname",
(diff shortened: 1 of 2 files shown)


javaducky (approved):
Thank you @cmergenthaler, this is awesome!

I was able to recreate the original issue where the starter pod was unable to start because all nodes had a runner deployed. Having applied the patch, the starter pod could run on any node having a runner pod assigned.

yorugac (approved):

Issue #118, opened 2022-06-07
init container


ranyhb:
hi,

i searched the crd and code and found out that the operator dosent have an option for an "initContainer".
i would like to know if you can add it in the next update/release.

yorugac:
Hi @ranyhb, thanks for the issue. Could you please describe your use case too?

BarthV:
Why not allow `k6 run ...` to directly accept url formated files (or archives if possible) ?

ranyhb:
I think you will need to configure the certs and many more other configurations.. 
Imagine also having some teams that use typescript instead of java script, before running the test to need to convert(build) the scripts in order to run them with k6

flashmann:
we would also love to see possibility to use init container to be able to clone git repo with test scripts

ranyhb:
I saw that you are planning to move to helm chart in the next version. In helm you can set hooks and their is a pre-install hook which work similarly to the init container so it can help us solve our problem.
 I would like to know when you are planning to release version 0.0.8.

robert-becker-hs:
Adding to the mix here, init containers would be very useful for us since we are loading our secrets inside an init container and making them available on a mount. So any initialization work would greatly benefit from an init container inside the CRD

andrei-trandafir:
Just trying to revive the conversation around this and show interest as well.

Init containers would be useful for our use-case as well as volume claims may seem a bit tricky for our situation. They'd allow us to mount scripts and data inputs that are stored in a separate bucket into the container. It'd be great to see this added sooner rather than later.

I know there's already a PR for this open, but it hasn't seen activity in over 3 months, so I'd be happy to take over this work to speed things up if possible.

PR #175, opened 2022-12-12
Add initializer spec

13013SwagR:


yorugac:
Thanks, @13013SwagR :slightly_smiling_face: 
Initializer was deliberately set to use the runner's spec because they're both running k6, unlike the starter. I.e. one needs an image with k6 for both initializer and runner while starter doesn't depend on k6, at least at the moment. So initializer was set to use runner's spec in order to minimize entities and keep configurations simple. Normally, a user of the operator shouldn't even bother with initializer, unless there are errors in the script and initializer terminates with an error. I think if we are to change this and put initializer into a separate spec, we should be more clear on why we're doing that.

So one case is with dapr, described in #164, but I'm not fully clear dapr usage really depends on a separate spec for initializer or if it was resolved with container specification fix. I'd like to confirm this first. Sadly, I didn't have time to test this myself yet.

Are there other cases?

PS this discussion is more fit for an issue and not a PR I guess :see_no_evil:

mcandeia:
> So one case is with dapr, described in #164, but I'm not fully clear dapr usage really depends on a separate spec for initializer or if it was resolved with container specification fix. I'd like to confirm this first. Sadly, I didn't have time to test this myself yet.

As I described there, "the problem" here is related on how kubernetes understand that a given job is completed. 

Since dapr is a sidecar (i.e one more container within the same pod) and the annotations for the runners and initializers are the same, the sidecar will be present in both, runners and initializers. The big difference is because when using runners I'm able to call the dapr shutdown [using its api](https://docs.dapr.io/operations/hosting/kubernetes/kubernetes-job/) ([see one example at dapr](https://github.com/dapr/dapr/blob/master/tests/perf/actor_id_scale/test.js#L57)) but as the initializers are totally out of my control I can't call shutdown, meaning that from the kubernetes point of view, the initializer job as a whole still pending even when the initializer container actually was successfully executed (because daprd still running). Furthermore, the operator leverages the kubernetes status to make sure the job is completed, see [here](https://github.com/grafana/k6-operator/blob/main/controllers/k6_initialize.go#L86).

Conclusion: The operator will not work when the initializer job has sidecar containers because the job isn't considered completed until all the containers are exited with code 0.

yorugac:
Thank you for the clarifications, @mcandeia!

> Furthermore, the operator leverages the kubernetes status to make sure the job is completed, see [here](https://github.com/grafana/k6-operator/blob/main/controllers/k6_initialize.go#L86).

This part is unlikely to change even with resolution of #138. So yes, some alternative is clearly needed.

@13013SwagR,

>  I think the initializer Job should really be just an init container on the Starter Pod which solve my issue and the dapr one for the initializer.

This pod is required for the k6 Cloud output support. It can't be merged with the starter: these are clear separate stages that must be executed sequentially.

I've tested this with Envoy / scuttle: there's actually another bug which was also reported recently in #179 - I added the fix in PR #182.
To sum up, enabled Envoy will work with `disableLogging: true`. Otherwise, initializer output will be polluted with sidecar's logs and cannot be processed. This is "expected behaviour" at the moment. This is true regardless of initializer job being separate or not.

As for request to make initializer job separate, we have three cases so far:
1) dapr sidecar shutdown
2) disabling Istio for initializer only but keeping it for runners
3) downloading the script in init container of initializer (also connected to #118)

Given that, I think it makes sense to make initializer job separate at this point of time. More complex logic would be to make initializer job run only for specific cases but that requires some more thinking, esp. given the limitation with the logs (described above with Envoy).

pkg/resources/jobs/initializer.go
```diff
@@ -27,28 +27,28 @@ func NewInitializerJob(k6 *v1alpha1.K6, argLine string) (*batchv1.Job, error) {
 		ports                        = append([]corev1.ContainerPort{{ContainerPort: 6565}}, k6.Spec.Ports...)
 	)
 
-	if k6.Spec.Runner.Image != "" {
-		image = k6.Spec.Runner.Image
+	if k6.Spec.Initializer.Image != "" {
+		image = k6.Spec.Initializer.Image
```
yorugac, line 31:
```suggestion
	if k6.Spec.Initializer == nil {
		k6.Spec.Initializer = k6.Spec.Runner.DeepCopy()
	}

	if k6.Spec.Initializer.Image != "" {
		image = k6.Spec.Initializer.Image
```

api/v1alpha1/k6_types.go
```diff
@@ -60,6 +60,7 @@ type K6Spec struct {
 	Separate    bool                   `json:"separate,omitempty"`
 	Arguments   string                 `json:"arguments,omitempty"`
 	Ports       []corev1.ContainerPort `json:"ports,omitempty"`
+	Initializer Pod                    `json:"initializer,omitempty"`
```
yorugac, line 63:
```suggestion
	Initializer *Pod                    `json:"initializer,omitempty"`
```

yorugac:
Hello @13013SwagR, This issue with initializer job got a higher internal priority lately. No pressure but could you please let me know if you'll be able to work on this PR in the coming weeks? Thank you!

13013SwagR:
Hey sorry for the late response, will update in the next days.

13013SwagR:
Hey @yorugac, 
Did the rebase and implemented your suggestions
Let me know if there's anything else
Thanks

yorugac (approved):
Thank you @13013SwagR, for coming back to this!

It looks OK to me; will be merging in :+1:

PR #177, opened 2023-01-04
Update README.md

emanuelef:
As of today the specified output for xk6 in the README is not working, I replaced it with xk6-prometheus-rw.


yorugac (approved):
Thanks for the fix, @emanuelef !

PR #217, opened 2023-04-24
Fix operator-sdk CSV and role auto generation

dkhachyan:
- Added new line after rbac  markers. Known issue: 
(https://github.com/kubernetes-sigs/controller-tools/issues/551)
- Added missing rbac markers
- Rename project to fix csv generator. 
- Operator bundle generation via `make bundle` requires `serviceAccountName` field in manager deployment manifest
https://github.com/operator-framework/operator-sdk/issues/4751#issuecomment-818930616

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -1,4 +1,3 @@
-
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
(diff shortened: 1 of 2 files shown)


yorugac:
Hi @dkhachyan! Thank you for the PR. Could you please describe what kind of issues you're trying to solve? We don't have this in our existing [backlog](https://github.com/grafana/k6-operator/issues) and though the links you added hint at the problems, a more verbose description would be highly appreciated in this case :slightly_smiling_face:

config/rbac/role.yaml
```diff
@@ -1,81 +1,91 @@
...
-    resources:
-      - secrets
-    verbs:
-      - list
-      - get
-      - watch
+- apiGroups:
+  - ""
+  resources:
+  - sercrets
```
yorugac, line 12: Is it possible to remove the change in tabs in RBAC list (in this file), so that we have a clean diff of what actually has changed and what has not? E.g. there's a typo being introduced here:
```suggestion
  - secrets
```
But it's not even supposed to be a change, i.e. it shouldn't be in a diff in the first place.

config/manager/manager.yaml
```diff
@@ -39,4 +39,9 @@ spec:
             requests:
               cpu: 100m
               memory: 50Mi
+          env:
+            - name: WATCH_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
```
yorugac, line 46: Please remove this change for `WATCH_NAMESPACE`. AFAIS, this will make k6-operator watch only one namespace for tests: while this may be a feature to add, it'd also be a breaking change for some setups. I suggest opening a feature request issue for this instead. Something like ~ "Restrict k6-operator to watch only one namespace in cluster".

yorugac (changes requested):
Hi @dkhachyan, thanks for the summary as well as for noticing these in the first place!
I have a couple of change requests for this PR: please see my comments, as well as a question around the image.

Also, the Yaml linter is failing: it'll need to be fixed as well :slightly_smiling_face:

config/manager/manager.yaml
yorugac: Thanks for removing namespace. Resolving.

yorugac (approved):
@dkhachyan Thanks for the updates. FYI, LGTM :+1: I'll be merging this PR tomorrow.

PR #225, opened 2023-05-04
Configurable watch namespaces

dkhachyan:


yorugac:
As mentioned in [PR 217](https://github.com/grafana/k6-operator/pull/217#discussion_r1183740038), this would be a breaking change for existing setups: that's why I suggested opening an issue instead. If namespace is to be configured, it must be optional, not mandatory by default.

dkhachyan:
It's not a breaking change. If WATCH_NAMESPACE is not set operator will watch all namespaces

config/manager/manager.yaml
```diff
@@ -39,4 +39,9 @@ spec:
             requests:
               cpu: 100m
               memory: 50Mi
+          env:
+            - name: WATCH_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
```
yorugac, line 46: Please remove changes in this file.

yorugac (changes requested):
Looking at this PR with fresh eyes, I think its possible to have an optional namespace if we keep just the changes in `main.go` and remove changes in manifest. I.e. by default `WATCH_NAMESPACE` is empty and the operator is watching everything. But if a user sets `WATCH_NAMESPACE` env var for the operator, only that namespace will be watched. Could you please adjust your PR accordingly?

config/manager/manager.yaml
dkhachyan: done

yorugac: Great; thanks for the update!

yorugac (approved):

artem-zherdiev-ingio:
Hey, any documentation on how to use it?

yorugac:
Documentation is here:
https://grafana.com/docs/k6/latest/set-up/set-up-distributed-k6/install-k6-operator/#watch-namespace

Issue #228, opened 2023-05-09
k6-operator test runs don't propagate maxMetricSamplesPerPackage setting


esquonk:
### Brief summary

Slack thread:
https://raintank-corp.slack.com/archives/C043CAPGCUR/p1683531983983669

ingest-api cloud service returns a config from create_test endpoint which includes maxMetricSamplesPerPackage setting among other things. These setting should be propagated to k6 instances so that it interacts with cloud output correctly.

One effect of this is that cloud ingest api gets larger-than-expected metrics packets that it cannot handle.

### k6-operator version or image

v0.0.9 

### K6 YAML

- "not available"

### Steps to reproduce the problem

Run a heavy test that generates lots of metric samples.

### Expected behaviour

Ingest packets are limited to maxMetricSamplesPerPackage.

### Actual behaviour

Ingest packets are not limited to maxMetricSamplesPerPackage.

yorugac:
This is considered resolved with PR #229

Issue #192, opened 2023-02-07
Metrics aggregation review


yorugac:
k6 have been using partial metrics aggregation for k6 Cloud processing, best explained here:
https://github.com/grafana/k6/blob/b750e206859a495b8a1fe12aa016dc8ed0e1c9db/cloudapi/config.go#L40-L96

The k6-operator cloud output run does not take this into account at the moment. Find out:
- did the new time series in k6 impact this partial aggregation?
- what is the most performant way to use cloud output in k6 at the moment?
- what needs to be changed in k6-operator invocation of k6 to take advantage of performant aggregation?

Implement changes as needs be.

yorugac:
Metrics aggregation can be turned on k6's side with these env vars:
```
  K6_CLOUD_AGGREGATION_MIN_SAMPLES
  K6_CLOUD_AGGREGATION_PERIOD
  K6_CLOUD_AGGREGATION_WAIT_PERIOD
  K6_CLOUD_METRIC_PUSH_INTERVAL
  K6_CLOUD_MAX_METRIC_SAMPLES_PER_PACKAGE
  K6_CLOUD_MAX_METRIC_PUSH_CONCURRENCY
```

But also watch out for the progress in https://github.com/grafana/k6/issues/2954: once it is resolved, aggregation may be turned on by default, without the need for any env vars tinkering.

cc @codebien

yorugac:
This is considered resolved with PR #229

PR #229, opened 2023-05-09
Set k6 env vars for metrics aggregation by default for cloud output

yorugac:
Fixing #228 

This branch with env vars for aggregation was already on my local so pushing it here.

Adding the env vars to the runners like this is a workaround which is not a good solution. This shouldn't be fixed on k6-operator side: it should be fixed on k6 side by making aggregation turned on by default, at least in case of `-out cloud` option.

@esquonk, please check if the numbers are correct: if they are, I'll merge it in the evening and make a release.

Related issue: #192

### DIFF
--- a/api/v1alpha1/conditions.go
+++ b/api/v1alpha1/conditions.go
@@ -148,6 +148,11 @@ func (k6status *K6Status) SetIfNewer(proposedStatus K6Status) (isNewer bool) {
 			isNewer = true
 		}
 		// log if proposedStatus.TestRunID is empty here?
+
+		// similarly with aggregation vars
+		if len(proposedStatus.AggregationVars) > 0 && len(k6status.AggregationVars) == 0 {
+			k6status.AggregationVars = proposedStatus.AggregationVars
+		}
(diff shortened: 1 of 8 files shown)


esquonk:
Ideally the numbers would be populated by the config received from `ingest-api`, not hardcoded. It is received in the same request as test run id, I wonder if it would be possible to access these numbers and propagate them to env vars?

https://github.com/grafana/k6/blob/master/cloudapi/config.go#L32

yorugac:
What API do you mean here? AFAIK, these env vars are set with hard-coded default values during k6 Cloud test runs.

https://github.com/grafana/k6/blob/master/cloudapi/config.go#L32

^ this is the env var, not an API.

yorugac:
Additional question, is config override going to differ for sequential test runs for the same org? @esquonk

yorugac:
With the latest addition, the config from `/v1/tests` is retrieved and passed to runners on each cloud output test. If config is ever not retrieved in this call, there will be an error and failed test.

PR #230, opened 2023-05-10
Add configurable readiness & liveness probe

0xF0D0:
Current k6 controller use default value of readiness & liveness probe.

In my case, this time slice is somewhat short and leads to readiness failure.

Through this pr, it can be configurable by k6 object


yorugac:
Hi @0xF0D0, thanks for the PR! I agree that the probes are rather simplistic now and should be improved. Passing the configured probes seems like one way to do that. But could you please clarify here:

> In my case, this time slice is somewhat short and leads to readiness failure.

what do you mean by the "time slice"?

0xF0D0:
I meant default readiness probe's `periodSeconds` and  `timeoutSeconds` are too small for my k6 workload

yorugac (changes requested):
@0xF0D0 I have a small change request to improve readability - please check it out. Otherwise, LGTM!

yorugac (approved):
Thanks for the update!

Issue #236, opened 2023-05-18
Implement logic for PLZ test runs


yorugac:
Subtasks:
- [X] Figure out resources conversion.
- [X] Unpacking test run data fully, including pre-signed URL thing.
- [X] The script is expected to be downloaded by URL. Consider options on how best to pass the script to the test run.
- [X] PLZ test run is a test run with cloud output but it is already created; so TestRunID field is set. That implies that CloudTestRun & CloudTestRunCreated should be set to True. Does such a test need other changes in logic? Confirm.
- [X] Add sending state info to the backend (on errors?) - ~~awaiting API~~
- [X] Polling state info from the backend and stopping test in case of abort.

yorugac:
Update: the last part (events) was added in PR https://github.com/grafana/k6-operator/pull/283

PR #239, opened 2023-05-31
Core implementation for test runs in PLZ mode

yorugac:
Issue: #236 

This builds up on top of PR #238


controllers/k6_controller.go
```diff
@@ -284,3 +348,46 @@ func (r *K6Reconciler) UpdateStatus(ctx context.Context, k6 *v1alpha1.K6, log lo
...
+
+	status, err := cloud.GetTestRunState(r.k6CloudClient, k6.Status.TestRunID, log)
+	if err != nil {
+		log.Error(err, "Failed to get test run state.")
+		return false
+	}
+
+	isAborted := status.Aborted()
+
+	// if isAborted {
```
javaducky, line 369: Nit: remove commented code.

controllers/k6_stopped_jobs.go
```diff
@@ -0,0 +1,130 @@
...
+
+	var deleteCount int
+
+	propagationPolicy := client.PropagationPolicy(metav1.DeletionPropagation(metav1.DeletePropagationBackground))
+	for _, job := range jl.Items {
+		if err = r.Delete(ctx, &job, propagationPolicy); err != nil {
+			log.Error(err, fmt.Sprintf("Failed to delete runner job %s", job.Name))
+			// do we need to retry here?
+		}
+		deleteCount++
```
javaducky, line 126: I don't think we want to increment the counter if the delete failed.

pkg/cloud/cloud_output.go
```diff
@@ -24,20 +25,34 @@ type TestRun struct {
...
+}
+
+func CreateTestRun(opts InspectOutput, instances int32, host, token string, log logr.Logger) (*cloudapi.CreateTestRunResponse, error) {
+	if client == nil {
+		client = NewClient(log, token, host)
+	}
+
+	cloudConfig := cloudapi.NewConfig()
+
+	if opts.External.Loadimpact.ProjectID > 0 {
```
javaducky, line 52: `Loadimpact`... 😢

pkg/cloud/plz.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"go.k6.io/k6/cloudapi"
+)
+
+const (
+	defaultApiUrl    = "https://api.k6.io"
+	defaultIngestUrl = "https://ingest.k6.io"
```
javaducky, line 12: I wonder how long the `*.k6.io` urls will be valid given future migration.

javaducky (commented):
As far as I can tell, things seem to be fine. Only a couple minor comments based upon quick review.

Not really sure how to fully test this functionality.

controllers/k6_stopped_jobs.go
yorugac: IIRC, the counter here is increased to avoid it being stuck (the return value of this function depends on `deleteCount`). I.e. it's an assumption that this error cannot be waited out - only logged.

But I'll check the possible errors from `Delete`. Maybe they can be further filtered out.

yorugac:
Branch updated and rebased with volumes as init container configuration as described in [this comment](https://github.com/grafana/k6-operator/issues/250#issuecomment-1652027200).

javaducky (approved):
🎉  LGTM!

PR #243, opened 2023-06-21
Fix 'no such file or directory' issue in initializer

vukor:
# Summary

This pr creates a directory where js test file is located to fix an issue when js test file is stored in volume (PVC/PV resources).

# How to reproduce

- create k6 CRD resource
```
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: test
spec:
  arguments: --vus 1 --iteration 1
  parallelism: 1
  runner:
    image: ...
  script:
    volumeClaim:
      file: some-dir/1/test.js
      name: k6-pvc
  scuttle:
    enabled: "false"
```

# Result

Initializer pod will be failed with `no such file or directory` error while trying to run:
```
k6 archive some-dir/1/test.js -O /tmp/some-dir/1/test.js.archived.tar ...
```
Because the `/tmp/some-dir/1/` directory doesn't exist


yorugac (approved):
Hi @vukor, thanks for catching this and fixing! LGTM :+1:

PR #247, opened 2023-07-12
Update image names to k6-operator over generic operator

### LINKED ISSUE #221: Image names should be more specific
### Feature Description

We're building all images for k6-operator under the prefix `ghcr.io/grafana/operator`. This might be confusing since Grafana has other operators (which are not hosted on `ghcr.io`, AFAIK, so there is no conflict yet). Still, it's best if we change the prefix for images to something like `ghcr.io/grafana/k6-operator`.

Once implemented, this change somewhat forces people to adjust their setups around the k6-operator and in their Dockerfiles if they're building from our runner image. So if someone has opinions to share, they're welcome to.

_No response_

_No response_

javaducky:
Fixes #221

### DIFF
--- a/pkg/resources/jobs/initializer.go
+++ b/pkg/resources/jobs/initializer.go
@@ -18,7 +18,7 @@ func NewInitializerJob(k6 *v1alpha1.K6, argLine string) (*batchv1.Job, error) {
 	}
 
 	var (
-		image                        = "ghcr.io/grafana/operator:latest-runner"
+		image                        = "ghcr.io/grafana/k6-operator:latest-runner"
 		annotations                  = make(map[string]string)
 		labels                       = newLabels(k6.Name)
 		serviceAccountName           = "default"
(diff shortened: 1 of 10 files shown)


pkg/resources/jobs/initializer.go
```diff
@@ -18,7 +18,7 @@ func NewInitializerJob(k6 *v1alpha1.K6, argLine string) (*batchv1.Job, error) {
 	}
 
 	var (
-		image                        = "ghcr.io/grafana/operator:latest-runner"
+		image                        = "ghcr.io/grafana/k6-operator:latest-runner"
```
yorugac, line 21: This change, along with similar ones in `*.go` files, breaks the operator until there's `latest` image is built. E2E test in Github Actions also is broken without it. So I've built a `new-pkgname` build from this branch manually, to have a `latest` image in `k6-operator` package. This way nobody should be affected by the change.

yorugac (approved):
@javaducky thanks for this change :+1:

PR #249, opened 2023-07-19
Add support for volume and volume mounts for runners

andrei-trandafir:
Currently there is no way to add different volume mounts to the runner pod, other than the default one that is created from the script spec. This PR adds support volumes and volume mounts.

The scenario we're aiming to utilise these in is when we're fetching the k6 script(s) and other artifacts from an external source (i.e. S3) through init containers, and copying them to the main containers using a Volume of type EmptyDir.


yorugac (changes requested):
@andrei-trandafir I have a change request: could you please remove changes to Makefile and instead add a new file, smth like `e2e/test-initcontainer-volumes.yaml`, which uses this new fields?

Otherwise, the changes seem OK to me and thanks for adding the Go test :slightly_smiling_face:

andrei-trandafir:
Thanks for the review @yorugac , i've reverted the Makefile changes and added a new sample k6 file in `e2e`.

e2e/test-initcontainer-volumes.yaml
```diff
@@ -0,0 +1,27 @@
...
+    volumeMounts:
+      - mountPath: /test
+        name: k6-provision-location
+    initContainers:
+      - image: busybox:1.28
+        command:
+          [
+            "sh",
+            "-c",
+            "cd /test; wget https://github.com/grafana/k6-operator/blob/main/e2e/test.js;",
```
yorugac, line 23:
```suggestion
            "cd /test; wget https://raw.githubusercontent.com/grafana/k6-operator/main/e2e/test.js;",
```
I think this link needs to be changed. Right now this sample test fails with:
```
time="2023-08-02T11:38:48Z" level=error msg="SyntaxError: file:///test/test.js: Unexpected token (7:0)\n   5 | \n   6 | \n>  7 | <!DOCTYPE html>\n     | ^\n   8 | <html lang=\"en\" data-color-mode=\"auto\" data-light-theme=\"light\" data-dark-theme=\"dark\"  data-a11y-animated-images=\"system\" data-a11y-link-underlines=\"false\">\n   9 |   <head>\n  10 |     <meta charset=\"utf-8\">\n\tat <internal/k6/compiler/lib/babel.min.js>:2:28536(99)\n\tat <internal/k6/compiler/lib/babel.min.js>:14:24413(11)\n\tat bound  (native)\n\tat s (<internal/k6/compiler/lib/babel.min.js>:1:1327(8))\n" hint="script exception"
```
So it downloads HTML instead of JS.

yorugac (changes requested):
@andrei-trandafir thanks for updates! The yaml test needs a very small update: it fails at the moment, please see the comment.

And one more thing. When trying to deploy from your branch as is, I hit an interesting error about byte limit:
```
The CustomResourceDefinition "k6s.k6.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```
It seems it can be resolved by adding parameter [`crd:maxDescLen=0` in Makefile](https://github.com/kubernetes-sigs/kubebuilder/issues/2556). But since you didn't add this parameter as part of PR, I wonder have you encountered this error during your testing?

andrei-trandafir:
@yorugac thanks for spotting the issue. I probably copied the wrong file 😅 

Regards the CRD install, I believe I encountered the issue once and didn't keep the solution in the Makefile in the end.
The way I solved it was by adding the following flag to the kubectl apply in the  `deploy` command in the Makefile:
```
--server-side=true
```
This removes the "last-applied-configuration" annotation from what I understand.

yorugac:
OK, I see. Actually, `--server-side=true` seems to work only on clean slate, i.e. without existing k6-operator deployment. Otherwise, I get these errors:
```
error: Apply failed with 9 conflicts: conflicts with "kubectl-client-side-apply" using apiextensions.k8s.io/v1beta1:
- .spec.validation.openAPIV3Schema.properties.spec.properties.initializer.properties.initContainers.items
- .spec.validation.openAPIV3Schema.properties.spec.properties.initializer.properties.volumeMounts.items
- .spec.validation.openAPIV3Schema.properties.spec.properties.initializer.properties.volumes.items
- .spec.validation.openAPIV3Schema.properties.spec.properties.runner.properties.initContainers.items
- .spec.validation.openAPIV3Schema.properties.spec.properties.runner.properties.volumeMounts.items
- .spec.validation.openAPIV3Schema.properties.spec.properties.runner.properties.volumes.items
- .spec.validation.openAPIV3Schema.properties.spec.properties.starter.properties.initContainers.items
- .spec.validation.openAPIV3Schema.properties.spec.properties.starter.properties.volumeMounts.items
- .spec.validation.openAPIV3Schema.properties.spec.properties.starter.properties.volumes.items
```
But also, for 'cleaner' solution I think it's better to change `make manifests` with `crd:maxDescLen=0` and not `make deploy` with  `--server-side=true`: so that there's an impact only on manifest generation during development and not on the bundle generation during deployments and releases. So, could you please add that `crd:maxDescLen=0` to `make manifests` as well? Assuming that option works in your setup as well :slightly_smiling_face:

andrei-trandafir:
For whatever reason, adding `crd:maxDescLen=0` results in the needed effect of not having descriptions, but changes the kind version to `apiextensions.k8s.io/v1beta1`, which results in the following error:
```
resource mapping not found for name: "k6s.k6.io" namespace: "" from "STDIN": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
```

I had changed the manifests command to:
```
manifests: controller-gen
	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```

andrei-trandafir:
Updating controlller-gen to v0.12.1 seems to have fixed it for me:
```
        ...
	go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.12.1 ;\
        ...
```
And then updating the CRD_OPTIONS to:
```
CRD_OPTIONS ?= "crd:maxDescLen=0"
```
As crd:trivialVersions is not an option anymore on that version and the default for crdVersions is `v1`.

Let me know how you'd like to proceed @yorugac. I'm happy to just change the version in the Makefile to what you suggested only if you think my case is an isolated case.

andrei-trandafir:
Added and tested the changes. Everything seems in order 👍🏼 
Thanks @yorugac

yorugac (approved):
LGTM! Seems to work AFAIS :+1: 
Thanks for the changes :smile:

Issue #206, opened 2023-03-17
Rename CRD fields to have standard naming


yorugac:
k6-operator has a non-standard (in context of Kubernetes API) naming for node selector: it's `nodeselector` instead of `nodeSelector`. It's preferable to fix it though it might be a breaking change for some.

yorugac:
Together with renaming, consider using the type [`NodeSelector`](https://github.com/kubernetes/api/blob/v0.26.2/core/v1/types.go#L2798) instead of `map[string]string`.

lplazas:
If this isn't assigned to anyone internally, I'd be happy to have it assigned to me.

JorTurFer:
I have drafted a PR fixing this

PR #274, opened 2023-08-21
fix: Use standard naming for nodeSelector definition

JorTurFer:
As it was exposed [here](https://github.com/grafana/k6-operator/issues/206), `nodeselector` uses wrong notation and breaks with Kubernetes standard naming. This PR renames the section to `nodeSelector`

**This is a breaking change**, if you want to maintain backward compatibility, I can update the PR adding support to both and a deprecation message on the wrong one

Fixes https://github.com/grafana/k6-operator/issues/206


yorugac (approved):
Hi @JorTurFer, thanks for the PR! LGTM :+1: 

In this case, I think let's change it as is, as breaking change, since it's basically a result of typo; I'd prefer not to support `nodeselector` long-term.

Issue #226, opened 2023-05-04
Rename K6 CRD to TestRun


yorugac:
In order to improve readability of our interfaces, we should rename the K6 CRD from:
```
kind: K6
```
To:
```
kind: TestRun
```

Why? `kind: K6` is not very descriptive as from the first glance, it might mean a lot of things. If we add new CRDs (see #191), we should try to remove the confusing names as well.

This is a breaking change for existing setups so consider supporting both names for some period of time.

yorugac:
`TestRun` CRD has been added as a duplicate of `K6` CRD with the latter being marked as deprecated in #286 
Missing dependency: #273 

We'll support both `K6` and `TestRun` for some time, to give people the opportunity to switch. But afterwards `K6` will be removed. IOW, we didn't get to the "breaking" part yet.

I'm leaving this issue open for now. What is left: 
- [ ] updating documentation
- [ ] removing `K6` CRD in the future™

yorugac:
`K6` CRD is removed in `main` branch (i.e. the next release) so closing this issue 🎉

jeromeguillaume:
Not sure if it's related to this change but (on January 24, 2024) the K6s operator doesn't work (looping restart) and it complains about following error:
`ERROR	controller-runtime.source.EventHandler	if kind is a CRD, it should be installed before calling Start	{"kind": "K6.k6.io", "error": "no matches for kind \"K6\" in version \"k6.io/v1alpha1\""}`

The image is: `ghcr.io/grafana/k6-operator:latest`

yorugac:
Hi @jeromeguillaume, yes, absolutely, it's related. `K6` has had a deprecation warning for more than a year, now it's fully removed and the `latest` will not work with it. The same goes to all future releases of k6-operator. Please switch to using `TestRun`. It's a one-line change, the API is exactly the same 🙂

yorugac:
@jeromeguillaume, that's very strange: v0.0.18 contains both `TestRun` and `K6` still. There was no release yet with this removal. At the moment, only `latest` image and kustomize setup contain the removal. 🤔 I guess this could happen if you're using the older installation that contains `K6` CRD with the `latest` image: that'd break indeed.

How are you installing / upgrading k6-operator?

yorugac:
The [v0.0.19 release](https://github.com/grafana/k6-operator/releases/tag/v0.0.19) is the first one that contains this change fully: bundle, Helm chart and `latest` images were all updated.

Issue #273, opened 2023-08-21
Upgrading the version of `controller-gen` used by k6-operator


bm1216:
## TO-DO
Update the version of controller-gen supported by `k6-operator`. 

## WHY?
`k6-operator` is currently supporting v0.3.0 of the `controller-gen` cli while v0.13.0 has been released. 
https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.13.0

Currently, `k6-operator` doesn't work with version v0.7.0 and above of `controller-gen`. This is due to addition of the `trivialVersions` option here - 
https://github.com/grafana/k6-operator/blob/e6001f213341a85b47da6c69f13136f4ac2a8c5e/Makefile#L21

which has been removed here in controller-gen - 
https://github.com/kubernetes-sigs/controller-tools/commit/09f1952580d4e40a3aab8b1c4f6bfc8de3ae0b0b#diff-fa737e39d27a85a80d2b211933392436e0e10b850c68cb53e3e8873b83d82071L42-L61

PR #286, opened 2023-09-13
Add a TestRun CRD as a duplicate of K6 CRD

yorugac:
Issue: #226 

With these changes, both `kind: K6` and `kind: TestRun` will be supported and will share the same logic. This is done as a transition period for everyone because K6 CRD will be removed in the future.


yorugac:
Additional changes in this PR include:
- `separate: false` for PLZ test runs
- PLZ CRD has now resources formatted according to Kubernetes conventions, expecting `resources.limits` to be defined. `resources.requests` can also be defined as part of PLZ (then they will be passed to each PLZ test run along with limits) but that is optional. 

cc @dgzlopes

yorugac:
Deprecation warning for K6 CRD requires resolving https://github.com/grafana/k6-operator/issues/273 at last.

PR #288, opened 2023-09-17
feat: use clusterip instead of DNS for service status check

CarlJi:


yorugac:
Hi @CarlJi, yes, this is one of the things that are constantly postponed because of other priorities :disappointed: Do you need an IPv6 support or only custom domain? You don't mention it in your description but it is present in code changes.

CarlJi:
> Hi @CarlJi, yes, this is one of the things that are constantly postponed because of other priorities 😞 Do you need an IPv6 support or only custom domain? You don't mention it in your description but it is present in code changes.

Actually for custom domain.

However, during my testing of the k6 operator in an IPv6 scenario, I also identified additional changes that need to be addressed, specifically like [curl_start](https://github.com/grafana/k6-operator/blob/4a09026463a5f02f0c1c54cb873e321fd0b3f14a/pkg/resources/containers/curl_start.go#L29), [curl_stop](https://github.com/grafana/k6-operator/blob/4a09026463a5f02f0c1c54cb873e321fd0b3f14a/pkg/resources/containers/curl_stop.go#L29) logic.

Supporting IPv6 may require a comprehensive design. I personally recommend prioritizing the resolution of the custom domain issue first.

yorugac:
@CarlJi, yes, agreed; that's why I asked :slightly_smiling_face: Let's do the following then: please simplify this PR down to domain changes and I'll merge that and then we'll see how that goes. My main worry here is not to break existing setups.

CarlJi:
@yorugac, updated now, please help review.

yorugac:
@CarlJi, thank you! Just letting you know: code-wise it seems fine, but I need to test this change, which would probably be next week. Sorry for the delay; too many things happening at once.

yorugac:
@CarlJi, sorry for the delay! I'm finally getting around to test this change. Could you please rebase your branch? There were dependency updates in main recently.

CarlJi:
> @CarlJi, sorry for the delay! I'm finally getting around to test this change. Could you please rebase your branch? There were dependency updates in main recently.

Done.

yorugac:
This change seems to hold up fine with E2E tests :raised_hands: The minimal way to test this particular case locally is with:
```yaml
# kind create cluster --config=./kind.custom.domain.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
kubeadmConfigPatches:
- |
  kind: ClusterConfiguration
  networking:
    dnsDomain: "cluster.foo"
```
I'll be using something similar :point_up: in automated tests.

yorugac (approved):

yorugac:
Thanks for working on this @CarlJi!

Issue #290, opened 2023-09-20
K6 is stuck on stage: `initialization` if the init job fails


JorTurFer:
### Brief summary

The operator creates the init job successfully, but if the pod fails for any reason, the operator doesn't notice it and the K6 job is stuck on the stage `initialization` until you manually remove it.

I'm willing to fix it (or at least to try it xD)

### k6-operator version or image

latest (sha256:79df77fea27ab5820ce3f25167268d5094be2fc10d182283fce9921e3786fed1)

### K6 YAML

Something that produces an error on init job

### Steps to reproduce the problem

Deploy a K6 manifest that produces a fail on the init pod. For example, linking a file that doesn't exist 

### Expected behaviour

The stage of K6 changes

### Actual behaviour

The stage of K6 doesn't change

yorugac:
Hi @JorTurFer! Thanks for working on this. I wouldn't call it a bug but more of an improvement in logic, TBH :smile: 

And this issue came up in several contexts recently! So linking key issues / PRs:
- #276 
- https://github.com/grafana/k6-operator/issues/260
- https://github.com/grafana/k6-operator/issues/222
- #283 

Quite a lot. I need to grok all of these to figure out what should be merged, changed, etc. It's in my TODO in the next couple of weeks so shouldn't be a long wait :+1: But as a heads up, there are duplicates and almost conflicts between the above.

JorTurFer:
> I wouldn't call it a bug but more of an improvement in logic

I have to disagree because literally the test is f**ked up. I mean, any kind of error during the initializing will stuck the test without any useful feedback (nor useless feedback, it doesn't give any feedback at all) 😄 
Any kind of automation over the test resource fails due to this. and it's quite annoying. Knowing the root cause, we have added more checks on different resources, but K6 resource's status isn't usable

JorTurFer:
Hello!
Any update about this topic?

yorugac:
Hi @JorTurFer, apologies for such a delay. Yes, actually, it's a good time to make this addition for the next release, given past and future work, but I'll have to ask for an update of your PR. Will comment over there.

alifemove:
Is this related to where it gets stuck like this
```
Init      [--------------------------------------] Starting outputs
default   [--------------------------------------]
```
and just does nothing after that?

My script is working locally, but when I try to run it in circleci this is all the further I get.

JorTurFer:
[...]

PR #291, opened 2023-09-20
fix: treat Failed as error during test init

JorTurFer:
Fixes https://github.com/grafana/k6-operator/issues/290

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"time"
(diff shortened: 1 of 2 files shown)


MrsDaehin:
is there a way to distinguish Test Errors ( for instance the ones from thresholds ) from Errors due to not being able to launch K6?

I think it would give more insights for troubleshooting.

JorTurFer:
> is there a way to distinguish Test Errors ( for instance the ones from thresholds ) from Errors due to not being able to launch K6?
> 
> I think it would give more insights for troubleshooting.

I think that we can add another state like `init-error` or `initialization-error` or similar to reflect this case. WDYT?

yorugac (changes requested):
It's been awhile so the PR will need to be rebased. But I've left comments to simplify the update :)

Thanks again!

irumaru:
Hello.
How far along are you now?
Is there anything I can do to help?

yorugac:
Closing this due to PR #401 being merged.

PR #295, opened 2023-09-26
chore: remove top newline from `config/crd` yaml

KaranbirSingh7:
#### What?
- kept running into the issue where copying _CRD_ directly from _k6-operator_ repo would always have an extra newline at top before YAML starts (_not sure if this is intentional, left for future comments maybe?!_) but this does cause a minor inconvenience when copying/pasting. 

#### Why?
- to allow copy/pasting without extra newline

### DIFF
--- a/config/crd/bases/k6.io_k6s.yaml
+++ b/config/crd/bases/k6.io_k6s.yaml
@@ -1,4 +1,3 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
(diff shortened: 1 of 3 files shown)


yorugac:
Hi @KaranbirSingh7, thanks for the PR but I'm afraid I cannot merge it. These manifests are auto-generated so they shouldn't be edited manually.

Good news is that the newline might disappear in future updates of dependencies. Specifically, related issue that will be worked on very soon: https://github.com/grafana/k6-operator/issues/273

KaranbirSingh7:
I wasn't aware of that. Fair point & thanks for clarifying @yorugac 🙏

PR #302, opened 2023-10-04
Remove hardcoded cluster domain

ausias-armesto:
This contribution allows to execute the `k6-operator` in clusters with different cluster name than `cluster.local`.
If the requirement is to have a full qualified name, a better approach would be to create an specific helm value for that, using the same pattern than bitnami charts [clusterDomain](https://github.com/bitnami/charts/blob/main/bitnami/elasticsearch/values.yaml#L50)


ausias-armesto:
It looks like it's already done. Thank you

Issue #307, opened 2023-10-11
Fails to start on Kubernetes 1.27


langesven:
### Brief summary

Hi,

we've recently been planning on giving the `k6-operator` a go so I set it up yesterday in our AWS EKS cluster (`v1.27.5-eks-43840fb`) but the manager fails to start with the below output:

```
kube-rbac-proxy I1011 06:18:25.661351       1 main.go:190] Valid token audiences: 
kube-rbac-proxy I1011 06:18:25.661433       1 main.go:262] Generating self signed cert as no cert is provided
kube-rbac-proxy I1011 06:18:26.427467       1 main.go:311] Starting TCP socket on 0.0.0.0:8443
kube-rbac-proxy I1011 06:18:26.427757       1 main.go:318] Listening securely on 0.0.0.0:8443
manager panic: runtime error: invalid memory address or nil pointer dereference
manager [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12cd7f0]
manager 
manager goroutine 1 [running]:
manager k8s.io/client-go/discovery.convertAPIResource(...)
manager     /go/pkg/mod/k8s.io/client-go@v0.26.1/discovery/aggregated_discovery.go:88
manager k8s.io/client-go/discovery.convertAPIGroup({{{0x0, 0x0}, {0x0, 0x0}}, {{0xc000870978, 0x15}, {0x0, 0x0}, {0x0, 0x0}, ...}, ...})
manager     /go/pkg/mod/k8s.io/client-go@v0.26.1/discovery/aggregated_discovery.go:69 +0x5f0
manager k8s.io/client-go/discovery.SplitGroupsAndResources({{{0xc000870018, 0x15}, {0xc0000444e0, 0x1b}}, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, ...}, ...})
manager     /go/pkg/mod/k8s.io/client-go@v0.26.1/discovery/aggregated_discovery.go:35 +0x2f8
manager k8s.io/client-go/discovery.(*DiscoveryClient).downloadAPIs(0x1832030?)
manager     /go/pkg/mod/k8s.io/client-go@v0.26.1/discovery/discovery_client.go:310 +0x478
manager k8s.io/client-go/discovery.(*DiscoveryClient).GroupsAndMaybeResources(0xc00051a1e0?)
manager     /go/pkg/mod/k8s.io/client-go@v0.26.1/discovery/discovery_client.go:198 +0x5c
[...]

PR #308, opened 2023-10-11
Update of Golang and dependencies

yorugac:
- [X] Update of Golang to 1.19
- [X] Adding new CI to run unit tests against later version of Kubernetes
- [X] Fixing newer version of Kubernetes to E2E with kind
- [X] Update of Kubernetes deps

Additionally, should fix https://github.com/grafana/k6-operator/issues/307


yorugac:
controller-runtime brings quite a few breaking changes in this update so some logic had to be re-wired. There are implications of those changes that we might be able to utilize but that'd require additional research.

Additionally, switching from `configmapsleases` to `leases` here, as it is the 2nd such deps update since https://github.com/grafana/k6-operator/pull/212.

PR #314, opened 2023-10-17
fix helm template indentation for deployment.env

shippo-eric:
This is a simple yaml typo fix.  Helm template indentation for `env` needs to be incremented.  

**before**:
```
          env:
          foo: bar
```

**after fix**:
```
          env:
            foo: bar
```

I take the liberty to update chart version, but feel free to let me know if that's not desired.


shippo-eric:
I made a mistake and realized `env` is a list of key/value pair, this is all good.

PR #320, opened 2023-10-23
ci: add golang lint and fixed. #157

### LINKED ISSUE #157: Improve linting
### Feature Description

Currently, k6-operator uses only `yaml-lint` for manifests. Add Golang linters.

k6 repo can be used as a main reference.

_No response_

u5surf:
Fixes #157 
Add golangci-lint on github actions ci and fixed some conventional issues of linting.

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,5 +1,5 @@
 ---
-name: "golangci-lint"
+name: "Golang Lint"
 on:
   - push
   - pull_request
@@ -16,5 +16,5 @@ jobs:
       - name: lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.54
-          args: --timeout=30m
+          version: v1.55
+          args: --timeout=5m
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,0 +1,20 @@
+---
+name: "golangci-lint"
+on:
+  - push
+  - pull_request
+
+jobs:
+  golangci:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.19'
+          cache: false
+      - name: lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.54
+          args: --timeout=30m
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,10 @@ fmt:
 vet:
 	go vet ./...
 
+# Run golangci-lint
+lint:
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.0
+	golangci-lint --timeout 5m run ./...
 # Generate code
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
--- a/controllers/testrun_controller.go
+++ b/controllers/testrun_controller.go
@@ -240,7 +240,7 @@ func (r *TestRunReconciler) reconcile(ctx context.Context, req ctrl.Request, log
 			// If TestRunRunning has just been updated, wait for a bit before
 			// acting, to avoid race condition between different reconcile loops.
 			t, _ := v1alpha1.LastUpdate(k6, v1alpha1.TestRunRunning)
-			if time.Now().Sub(t) < 5*time.Second {
+			if time.Since(t) < 5*time.Second {
 				return ctrl.Result{RequeueAfter: time.Second * 2}, nil
 			}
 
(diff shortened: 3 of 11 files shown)


controllers/testrun_controller.go
```diff
@@ -268,7 +268,7 @@ func (r *TestRunReconciler) reconcile(ctx context.Context, req ctrl.Request, log
 		// delete if configured
 		if k6.GetSpec().Cleanup == "post" {
 			log.Info("Cleaning up all resources")
-			r.Delete(ctx, k6)
+			r.Delete(ctx, k6) //nolint:errcheck
```
yorugac, line 271: Can this linter be fixed with smth like:
```
_ = r.Delete(ctx, k6)
```
instead of exception?

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.19'
+          cache: false
+      - name: lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.54
```
yorugac, line 19: The latest is `v1.55.0` according to [official page](https://github.com/golangci/golangci-lint/releases). They update pretty often. Can we try to go with `v1.55`?

pkg/types/k6cli.go
```diff
@@ -14,12 +14,10 @@ type CLI struct {
 
 func ParseCLI(arguments string) *CLI {
 	lastArgV := func(start int, args []string) (end int) {
-		var nextArg bool
 		end = start
-		for !nextArg && end < len(args) {
+		for end < len(args) {
```
yorugac, line 18: I don't remember why it was written with `nextArg` :sweat_smile: But the change seems to be correct and the unit test is OK... For my understanding, what was the linter that made you add this change?

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: "golangci-lint"
```
yorugac, line 2: Please change:
```suggestion
name: "Golang Lint"
```
I've been trying to keep unified naming in CIs lately.

yorugac (changes requested):
Thanks for working on this, @u5surf! For starters, it's probably fine to go with the default linting as here: we can always add additional rules later :+1: 
I have a few questions and small CRs - please check the comments.

yorugac:
One more thing: can you please add a small instruction / command on how to run this linter locally? It could be a target in Makefile or part of internal docs (i.e. `docs/` folder). It can be very frustrating to fix linting in Github Actions so running it quickly locally is preferable.

controllers/testrun_controller.go
u5surf: @yorugac 
We also should be fixed like that below?
```
diff --git a/main.go b/main.go
index 169614b..ed60b2f 100644
--- a/main.go
+++ b/main.go
@@ -89,8 +89,8 @@ func main() {
                os.Exit(1)
        }

-       mgr.AddHealthzCheck("health", healthz.Ping)
-       mgr.AddReadyzCheck("ready", healthz.Ping)
+       mgr.AddHealthzCheck("health", healthz.Ping) //nolint:errcheck
+       mgr.AddReadyzCheck("ready", healthz.Ping)   //nolint:errcheck

        if err = (controllers.NewK6Reconciler(&controllers.TestRunReconciler{
                Client: mgr.GetClient(),
```

pkg/types/k6cli.go
u5surf:
```
pkg/types/k6cli.go:22:5: ineffectual assignment to nextArg (ineffassign)
				nextArg = true
				^
```

yorugac: Got it, thanks!

controllers/testrun_controller.go
yorugac: If the linter allows it, yes please!

Makefile
```diff
@@ -98,6 +98,10 @@ fmt:
 vet:
 	go vet ./...
 
+# Run golangci-lint
+lint:
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.0
+	golangci-lint --timeout 30m run ./...
```
yorugac, line 104: Nice :raised_hands:  Timeout is pretty large: our code base is rather small so shouldn't need that much in theory :thinking:  Either way, this would work, thanks for adding Makefile!

u5surf: @yorugac 
I could reduce the timeout.

controllers/testrun_controller.go
u5surf: @yorugac 
I also fixed these.

Makefile
yorugac: :+1: thanks. It should be fine with 5m, and if not, we can always tinker with it.

yorugac (approved):
LGTM! Thank you @u5surf, for working on this :smile:

PR #326, opened 2023-10-30
Enable custom pod labels for helm chart

MarkDPierce:
Hey, we leverage tooling that requires certain labels to be present. While deploying into our ecosystem I noticed the pods don't have the capability to have custom labels. This PR is meant to include that annotation and feature for others that require specific pods labels

### DIFF
--- a/charts/k6-operator/Chart.yaml
+++ b/charts/k6-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.0.11"
 description: A Helm chart to install the k6-operator
 name: k6-operator
-version: 1.2.0
+version: 2.0.0
 kubeVersion: ">=1.16.0-0"
 home: https://k6.io
 sources:
--- a/charts/k6-operator/samples/customAnnotationsAndLabels.yaml
+++ b/charts/k6-operator/samples/customAnnotationsAndLabels.yaml
@@ -4,6 +4,10 @@ customAnnotations:
 customLabels:
   "customized-labels": "k6-operator"
 
+podLabels:
+  environment: production
+  owner: development
+
 certManager:
   enabled: true
 
--- a/charts/k6-operator/templates/_helpers.tpl
+++ b/charts/k6-operator/templates/_helpers.tpl
@@ -72,6 +72,14 @@ Create the name of the service account to use
   {{- end }}
 {{- end -}}
 
+{{- define "k6-operator.podLabels" -}}
+  {{- if .Values.podLabels }}
+    {{- with .Values.podLabels }}
+      {{- toYaml . }}
+    {{- end }}
+  {{- end }}
+{{- end -}}
+
 {{- define "k6-operator.customAnnotations" -}}
   {{- if .Values.customAnnotations }}
     {{- with .Values.customAnnotations }}
--- a/charts/k6-operator/templates/deployment.yaml
+++ b/charts/k6-operator/templates/deployment.yaml
@@ -19,6 +19,7 @@ spec:
     metadata:
       labels:
         {{- include "k6-operator.selectorLabels" . | nindent 8 }}
+        {{- include "k6-operator.podLabels" . | nindent 8 }}
     spec:
       containers:
         {{- if .Values.authProxy.enabled }}
--- a/charts/k6-operator/values.yaml
+++ b/charts/k6-operator/values.yaml
@@ -4,6 +4,9 @@ customAnnotations: {}
 # customLabels -- Custom Label to be applied on all resources
 customLabels: {}
 
+# podLabels -- Custom Label to be applied on all pods
+podLabels: {}
+
 # nodeSelector -- Node Selector to be applied on all containers
 nodeSelector: {}
 


yorugac:
Hi @MarkDPierce, why `customLabels` cannot be used? They should be applied to Deployment as well, [here](https://github.com/grafana/k6-operator/blob/7ce63b1790f30f6b9693e8a4a2f0fed0bb47c2ac/charts/k6-operator/templates/deployment.yaml#L10).

MarkDPierce:
When I use a custom label of `environment: test` and generate the manifest I get the following.

```
# Source: k6-operator/charts/k6-operator/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: release-name-k6-operator-controller-manager
  namespace: release-name-system
  labels:
    control-plane: "controller-manager"
    app.kubernetes.io/component: controller
    helm.sh/chart: k6-operator-1.2.0
    app.kubernetes.io/name: k6-operator
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "0.0.11"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/part-of: k6-operator
    environment: test
  annotations:

spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: k6-operator
      app.kubernetes.io/instance: release-name
  template:
    metadata:
      labels:
        app.kubernetes.io/name: k6-operator
        app.kubernetes.io/instance: release-name
    spec:
      containers:
        - name: kube-rbac-proxy
          image: "europe-docker.pkg.dev/dbk-ecom-cicd/docker-hosted/kubebuilder/kube-rbac-proxy:v0.8.0@sha256:34e8724e0f47e31eb2ec3279ac398b657db5f60f167426ee73138e2e84af6486"
          imagePullPolicy: IfNotPresent
          args:
            - --secure-listen-address=0.0.0.0:8443
            - --upstream=http://127.0.0.1:8080/
            - --logtostderr=true
            - --v=10
          ports:
            - containerPort: 8443
              name: https
        - name: manager
          image: "europe-docker.pkg.dev/dbk-ecom-cicd/docker-hosted/k6-operator/controller:v0.0.11@sha256:d80c85a54d3687b346cba8e44421364f87d54f89b6a7adace839d49e0b50c6d8"
          imagePullPolicy: Always
          resources:
            limits:
              cpu: 50m
              memory: 50Mi
            requests:
              cpu: 50m
              memory: 50Mi
          command:
            - /manager
          args:
            - --enable-leader-election
            - --metrics-addr=127.0.0.1:8080
      serviceAccount: k6-operator-controller
      terminationGracePeriodSeconds: 10
```
These customLabels are not included in the template metadata. Due to security reasons we need labels at the pod level. So every templated pod will get these custom labels.

Thanks for actually asking because I totally forgot to add `{{- include "k6-operator.customLabels" . | nindent 8 }}` to the deployment.yaml template.

MarkDPierce:
Sorry, that was a long winded explanation to just say deployment.yaml only has 
```
spec:
  replicas: 1
  selector:
    matchLabels:
      {{- include "k6-operator.selectorLabels" . | nindent 6 }}
  template:
    metadata:
      labels:
        {{- include "k6-operator.selectorLabels" . | nindent 8 }}
    spec:
```
So customLabels wouldnt work in this scenario.

yorugac:
Thanks for full explanation. Now after some looking around, I'm not fully certain we should have `customLabels` at all :confused:  E.g. these charts don't seem to propagate user-defined labels at the level of Deployments or StatefulSets:
https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/values.yaml
https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml

In k6-operator case, there's only one Deployment which makes it easier; though I suspect in the future we might need to make further breaking changes here.

I'd appreciate some feedback from community here, about how everyone uses labels in their charts :slightly_smiling_face:

charts/k6-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.0.11"
 description: A Helm chart to install the k6-operator
 name: k6-operator
-version: 1.2.0
+version: 2.0.0
```
yorugac, line 5: Why do you think it should be a breaking change? This seems like a "pure" addition: old deployments should continue to work as is. I.e. why not 1.3.0?

FYI, I'll need to test it a bit and probably return to this PR ~ tomorrow.

MarkDPierce:
It's a feature change and other charts we've worked on within the open
source community prefer a major bump. I also attempted to find any docs you
might have had towards the process. So went with best guess, I'm happy to
change or conform to what ever you deem appropriate.

charts/k6-operator/Chart.yaml
yorugac: Thanks for sharing your experience with versioning of other Helm charts. The chart in k6-operator was added very recently and we're figuring out the "right way" still :sweat_smile: 
Let's go with major bump this time indeed :+1:

yorugac (changes requested):
Sorry for the delay! 
The chart generated looks fine and working :raised_hands: I have a small change request: could you please add `podLabels` to example in `charts/k6-operator/samples/customAnnotationsAndLabels.yaml`? It could be an additional sample (file) too, btw.

And another thing. I've noticed that there is an additional empty line in manifest when one omits `podLabels` like this:

```yaml
  ...
    metadata:
      labels:
        app.kubernetes.io/name: k6-operator
        app.kubernetes.io/instance: release-name
        
    spec:
      ...
```
:confused: The same empty line is inserted with custom labels which use similar definition and default `{}` so it's not just change in this PR. This is a minor thing but if you or someone knows how to fix it, I'd appreciate a PR here :smile:

MarkDPierce:
Hey sorry for the delay. Work had prio.

I took a stab at the line breaks in your labels and annotations but it might require a bit of refactoring. At least with both you have `| default "" |` which means it's by design. So I took a stab at a bit of refactoring and managed to achieve the following
```yaml
---
# Source: k6-operator/templates/namespace.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: release-name-system
  labels:
    app.kubernetes.io/name: release-name-k6-operator
    control-plane: "controller-manager"
    foo: bar
  annotations:
    bar: foo
---
# Source: k6-operator/templates/serviceAccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: k6-operator-controller
  namespace: release-name-system
  labels:
    app.kubernetes.io/component: controller
    helm.sh/chart: k6-operator-2.0.0
    app.kubernetes.io/name: k6-operator
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "0.0.11"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/part-of: k6-operator
    foo: bar
  annotations:
    bar: foo
```
But I do think the changes would be out of scope for this change.

yorugac (approved):

yorugac:
@MarkDPierce yes, the line break would be out of scope, but thanks for taking a look. If you see the solution, please feel free to open a PR :slightly_smiling_face: ATM, I don't think `default ""` is a must so long as the values are passed correctly or are omitted correctly :thinking:

Either way, the `podLabels` change looks good, thank you for this!

Issue #328, opened 2023-11-02
Allow multiple replicas to be configured via helm chart


josh-ferrell:
### Feature Description

Allow multiple manager replicas to be configured via helm chart for redundancy.

yorugac:
Hi @josh-ferrell, thanks for opening this. 
AFAIR, the issue of multiple replicas wasn't raised before. k6-operator does not support multiple replicas at the moment, regardless of what installation method you use. Which in practice means that some use cases would work with multiple replicas deployed and some will not.
So independently of Helm, could you please describe the use case for having more than 1 replica?

josh-ferrell:
Having multiples running across zones for resiliency is a policy of ours. I assumed with the leader election that a second wouldn't be reconciling the CRs until they became the leader. TBH I'm not very familiar with k6 yet but if you can advise as to where multiple replicas would crash I can take a look at addressing them.

yorugac:
@josh-ferrell I'm reviewing multiple replicas setup again and I believe you're right and we won't need to change anything. TBH, the last time I checked multiple replicas workflow was almost a year ago and leases just weren't present then (in our deps). But now it looks like yes, lease solves the issues: only one replica will reconcile the resources while the rest remain on standby.

FYI, I'll continue testing this and will come back to you (and PR) a bit later this week. And thanks again for bringing this up to attention! :smile:

yorugac:
Closed with https://github.com/grafana/k6-operator/pull/329

PR #329, opened 2023-11-02
Allow manager replicas to be configured via helm chart

josh-ferrell:
Allow the number of manager replicas to be set via helm chart values. Enables #328


josh-ferrell:
I tested this with manager.replicas set to two. Lease holder was correct so leader election seems to be working. Pods both became ready and stayed healthy.

josh-ferrell:
Successfully ran e2e tests

charts/k6-operator/values.yaml
```diff
@@ -39,6 +39,8 @@ authProxy:
   readinessProbe: {}
 
 manager:
+  # manager.replicaCount -- number of controller-manager replicas to run (default: 1)
+  replicaCount: 1
```
yorugac, line 43: I've just now noticed this: can we go with `replicas` here? It's under `manager` so should be clear what it is and it's what's used in Kubernetes definition.

yorugac (changes requested):
@josh-ferrell I'm happy to merge this change but there are a couple of small change requests: I suggest to change the name to `replicas` and if it's not an issue, please bump the [Chart version](https://github.com/grafana/k6-operator/blob/main/charts/k6-operator/Chart.yaml) to 3.0.0 - it's what triggers the release.
I'm going to add that last bit about versions to [Contributing guide](https://github.com/grafana/k6-operator/issues/330) too.

charts/k6-operator/values.yaml
josh-ferrell: Got it, done

josh-ferrell: Sounds good, done.

yorugac (approved):
LGTM. Thank you @josh-ferrell!

PR #335, opened 2023-11-10
fix issue with service monitor

### LINKED ISSUE #324: Helm chart: ServiceMonitor namespace is not templated (system)
Trying to install k6-operator via Helm and prometheus.enabled=true raises an error about "namespace system not found".
Inspecting the helm files, serviceMonitor.yaml has an hardcodded value for the ServiceMonitor namespace.

https://github.com/grafana/k6-operator/blame/0c3bf71105f8ce298a9801c436841802a5995545/charts/k6-operator/templates/prometheus/serviceMonitor.yaml#L15

edwintye:
Fixes #324 where helm install with flag `--set prometheus.enabled=true` fails unless the `system` namespace exists.  Set this to the same namespace as all the other resources installed via the chart. Additionally, restrict the namespace selector to be within the same namespace as well so there is chance of accidental cross namespace lookup and drag in another service by chance.

Furthermore, fixes the selector label on the service itself as the current select 

```yaml
selector:
  control-plane: "controller-manager"
```

points to nothing. This can be easily verified without the need of a `ServiceMonitor` by port forwarding to the service and perform a simply curl against the endpoint. The change is to use the `selectorLabels` that is used under `.spec.template.metadata.labels` in the `Deployment`.

### DIFF
--- a/charts/k6-operator/templates/prometheus/serviceMonitor.yaml
+++ b/charts/k6-operator/templates/prometheus/serviceMonitor.yaml
@@ -12,8 +12,11 @@ metadata:
   annotations:
     {{- include "k6-operator.customAnnotations" . | default "" | nindent 4 }}
   name: controller-manager-metrics-monitor
-  namespace: system
+  namespace: {{- include "k6-operator.namespace" . }}
 spec:
+  namespaceSelector:
+    matchNames:
+      - {{- include "k6-operator.namespace" . }}
   endpoints:
     - path: /metrics
       port: https
--- a/charts/k6-operator/templates/deployment.yaml
+++ b/charts/k6-operator/templates/deployment.yaml
@@ -18,6 +18,7 @@ spec:
   template:
     metadata:
       labels:
+        control-plane: "controller-manager"
         {{- include "k6-operator.selectorLabels" . | nindent 8 }}
         {{- include "k6-operator.podLabels" . | nindent 8 }}
     spec:


charts/k6-operator/templates/service.yaml
```diff
@@ -17,5 +17,5 @@ spec:
     port: 8443
     targetPort: https
   selector:
-    control-plane: "controller-manager"
+    {{- include "k6-operator.selectorLabels" . | nindent 4 }}
```
yorugac, line 20: I suggest to fix the Chart so that `control-plane: "controller-manager"` would work. Rationale: it'd be preferable to keep the labeling consistent with kustomize base, i.e. here:
https://github.com/grafana/k6-operator/blob/main/config/manager/manager.yaml

charts/k6-operator/templates/prometheus/serviceMonitor.yaml
```diff
@@ -12,8 +12,11 @@ metadata:
   annotations:
     {{- include "k6-operator.customAnnotations" . | default "" | nindent 4 }}
   name: controller-manager-metrics-monitor
-  namespace: system
+  namespace: {{- include "k6-operator.namespace" . }}
 spec:
+  namespaceSelector:
```
yorugac, line 17: It [seems like](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#namespace-selector) `namespaceSelector` requires 1.24+ ? We're "officially" 1.23+ now (can be seen [here](https://grafana.com/docs/grafana-cloud/k6/author-run/private-load-zone-v2/#before-you-begin)). I think it'd be nice to increase this version some time ~ next year though, and then this addition would be great.

yorugac (changes requested):
Hi @edwintye, thank you for working on this! Namespace fix is great and working :+1: There are a couple of change requests though: please see the comments.

charts/k6-operator/templates/prometheus/serviceMonitor.yaml
edwintye: `namespaceSelector` is part of the [ServiceMonitor CRD](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#servicemonitor) and not part of native k8s semantics. From my perspective, whether the field is usable in an earlier version of k8s is probably an issue for the prometheus operator (or any operator which implements the CRD) and not the concern here.

charts/k6-operator/templates/service.yaml
edwintye: Sure, whatever changes you guys want. I have moved the label inside the deployment template instead.

charts/k6-operator/templates/prometheus/serviceMonitor.yaml
illrill: @edwintye is right, the above [link](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#namespace-selector) is specific to Pods, not ServiceMonitors.

yorugac: Thanks for clarification and the link! Agreed with the logic of this being the issue at the level of Prometheus operator :+1:

yorugac (approved):
Thank you for the updates @edwintye. LGTM :+1:

datsabk:
@yorugac - What is the target date to release this change in the chart?

yorugac:
@datsabk, this change is over here: https://github.com/grafana/helm-charts/releases/tag/helm-k6-operator-3.2.0

PR #338, opened 2023-11-14
docs: add kustomize to prerequisites

aslafy-z:
kustomize is needed for `make bundle`.


yorugac (approved):
Thanks for noticing and fixing @aslafy-z.

PR #340, opened 2023-11-20
Update _helpers.tpl

shahid430:
Changed namespace from default to monitoring

### DIFF
--- a/charts/k6-operator/templates/_helpers.tpl
+++ b/charts/k6-operator/templates/_helpers.tpl
@@ -89,7 +89,7 @@ Create the name of the service account to use
 {{- end -}}
 
 {{- define "k6-operator.namespace" -}}
-  {{- if eq .Release.Namespace "default" }}
+  {{- if eq .Release.Namespace "monitoring" }}
     {{- printf "%v-system" .Release.Name | indent 1 }}
   {{- else }}
     {{- .Release.Namespace | indent 1 }}


charts/k6-operator/templates/_helpers.tpl
```diff
@@ -89,7 +89,7 @@ Create the name of the service account to use
 {{- end -}}
 
 {{- define "k6-operator.namespace" -}}
-  {{- if eq .Release.Namespace "default" }}
+  {{- if eq .Release.Namespace "monitoring" }}
```
illrill, line 92: Looks to me that this function is designed to default the installation into `{{ .Release.Name }}-system` instead of Helms default choice of the `default` namespace, in the scenario that the user doesn't specify any namespace using `helm install --namespace`.

This change would mean that if a user specifies `monitoring` as the desired namespace, the installation goes into `{{ .Release.Name }}-system`. In contrast, if a user doesn't specify any namespace, the installation goes into `default` (which is what this function is designed to prevent).

yorugac:
@shahid430, thanks for the PR but it's indeed not clear what this fix is for: unless specified, namespace will be `%v-system` as is defined here:
https://github.com/grafana/k6-operator/blob/main/charts/k6-operator/templates/_helpers.tpl#L93

@illrill Thanks for helping out :smile:

yorugac:
@shahid430 I'm closing this PR due to reasons stated above. Please feel free to re-open / open a new one if there's something else!

PR #345, opened 2023-12-06
feat: container level security context

### LINKED ISSUE #297: Add in runner and starter possibility to configure allowPrivilegeEscalation and capabilities in container SecurityContext
### Feature Description

I tried to start following kind of test in our cluster:

`
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6-sample
  namespace: test
spec:
  parallelism: 4
  script:
    configMap:
      name: stress-test
      file: test.js
`

but this fails to work because it gets rejected by our policy engine, which requires the workloads to e.g. run as non-root and set seccompProfile. I noticed that it is possible to configure security contexts for the runner and starter so I tried something like this:

`
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6-sample
  namespace: test
spec:
  parallelism: 4
  script:
    configMap:
      name: stress-test
      file: test.js
  runner:
    securityContext:
      runAsNonRoot: true
      runAsUser: 1000
      seccompProfile:
        type: RuntimeDefault
      allowPrivilegeEscalation: false
      capabilities:
        drop:
          - ALL
  starter:
    securityContext:
      runAsNonRoot: true
      runAsUser: 1000
      seccompProfile:
        type: RuntimeDefault
      allowPrivilegeEscalation: false
      capabilities:
        drop:
          - ALL
`

otherwise it works, but unfortunately the operator does not seem to support "allowPrivilegeEscalation" and "capabilities"-configurations i.e. I get following error, already while applying the manifest:

`
error: error validating "k6.yml": error validating data: [ValidationError(K6.spec.runner.securityContext): unknown field "allowPrivilegeEscalation" in io.k6.v1alpha1.K6.spec.runner.securityContext, ValidationError(K6.spec.runner.securityContext): unknown field "capabilities" in io.k6.v1alpha1.K6.spec.runner.securityContext, ValidationError(K6.spec.starter.securityContext): unknown field "allowPrivilegeEscalation" in io.k6.v1alpha1.K6.spec.starter.securityContext, ValidationError(K6.spec.starter.securityContext): unknown field "capabilities" in io.k6.v1alpha1.K6.spec.starter.securityContext]; if you choose to ignore these errors, turn validation off with --validate=false
`

I tried to find out if this could be updated in the operator-code, and it seemed that the issue is related to this:
https://github.com/grafana/k6-operator/blob/main/api/v1alpha1/k6_types.go#L40

i.e. corev1.PodSecurityContext is being used, which does not support these configurations since they should get configured actually at the container level.

For now I can get around this issue by e.g. defining policy exceptions for this specific case, but of course it would be much nicer if you could configure this already in the k6-kind.

If you have an idea how this could be achieved, I would be more than happy to test this out in our environment.

_No response_

_No response_

llaszkie:
Hi!

The feature differentiate securityContext on the pod and container level. It makes the 'k6-operator` capable to operate on K8S >= 1.25 when PSS (Pod Security Standards) the built-in Pod Security Admission controller is in place and level is set to **Restricted**.

Sample usage:

```yaml
apiVersion: k6.io/v1alpha1
kind: TestRun
spec:
  runner:
   securityContext:
     runAsNonRoot: true
   containerSecurityContext:
     allowPrivilegeEscalation: false
```


yorugac:
Hi @llaszkie, thanks on working on this!
FYI, I'll be able to review this PR fully on the next week. For now, just one question: you chose to add a new field to Pod. 
Kubernetes does have 2 separate objects but since our TestRun API doesn't expose `containers`, a new field has to be added with a totally new name `containerSecurityContext`. Have you considered making a wrapper struct `securityContext` instead of a new field that will contain fields from both objects and then can be reduced to each with `.ToPodSecurityContext()` and `.ToContainerSecurityContext()` methods? This is the approach that I wanted to explore in context of this issue but didn't have time to dig into details yet: hence, the question :slightly_smiling_face:

llaszkie:
Hi @yorugac ,

Yes - I was considering the wrapper solution - as suggested in the #297. But it could bring more challenges than benefits. In the K8S definitions those contexts are defined separately ([pod here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podsecuritycontext-v1-core and [container here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#securitycontext-v1-core)), but they overlap in some attributes (sic!), like `seccompProfile`. And it is a valid use case (for example for pod volumes) - the overlap handling is defined in spec too. How then you would distinguish the user intention to set the one on the Pod, but not on the Container in your API exposing only combined wrapper? 

What could be done instead is to explicitly expose Container in your API. It would even more align both models (K8S and K6) and solution could be cleaner? Let me know how about that approach.

yorugac:
> What could be done instead is to explicitly expose Container in your API

WDYM by that?

After some looking into this, yes, I agree that there's obviously a goal of exposing a more granular control over security context in Kubernetes definitions, hence two objects. It'd be nice if we could avoid over-complicating our API. I think the main problem with the current implementation is that `containerSecurityContext` is being passed only to the main container and not to init containers. It probably should be passed everywhere if it is defined.

yorugac:
> init containers

I.e. here: https://github.com/grafana/k6-operator/blob/main/pkg/resources/jobs/helpers.go#L120-L147

llaszkie:
> What could be done instead is to explicitly expose Container in your API

I was thinking about extracting the container settings to a separate object embedded within the Pod. Probably too much overhead 😄 

Thanks for the hint for the init containers - absolutely right. Fixed.

yorugac:
@llaszkie, apologies for the delay. Thanks for the update; the changes seem to be fine :+1: But could you please rebase the branch against `main`? It can't be merged ATM.

llaszkie:
Good morning! @yorugac: rebased. I suppose we are good to go now :-)

charts/k6-operator/README.md
```diff
@@ -1,6 +1,6 @@
 # k6-operator
 
-![Version: 3.1.0](https://img.shields.io/badge/Version-3.1.0-informational?style=flat-square) ![AppVersion: 0.0.11](https://img.shields.io/badge/AppVersion-0.0.11-informational?style=flat-square)
+![Version: 3.3.0](https://img.shields.io/badge/Version-3.3.0-informational?style=flat-square) ![AppVersion: 0.0.11](https://img.shields.io/badge/AppVersion-0.0.11-informational?style=flat-square)
```
yorugac, line 3: Please update `charts/k6-operator/Chart.yaml` instead and run `helm-docs` command. This Readme is not supposed to be edited manually.

yorugac (changes requested):
@llaszkie, I have to ask for 2 more things though both should be pretty quick:
- please see my comment about Helm chart (I accidentally posted it as a comment instead of including into review)
- it'd be nice to add a sample of `TestRun` definition with a new option into `config/samples`.

charts/k6-operator/README.md
yorugac: @llaszkie, thank you but you need to git commit the `charts/k6-operator/Chart.yaml` too - it is a tracked file :sweat_smile:

yorugac (approved):
Thank you for the updates, @llaszkie!

PR #351, opened 2023-12-08
feat: add initializer envFrom field

### LINKED ISSUE #348: The initializer pod spec does not have envFrom
### Brief summary

Hi, 
I am pretty new to k6 and the operator usage. The `initializer` pod spec does not seem to have the `envFrom` option and just has `env` option as can be seen from [these lines](https://github.com/grafana/k6-operator/blob/main/pkg/resources/jobs/initializer.go#L118-L127) compared to the `runner` pod spec which does have it. 

The JSON spec for the `initializer` does seem to expose the `envFrom` though as can be inspected from running `kubectl explain TestRun.spec.initializer.envFrom`: 

```
GROUP:      k6.io
KIND:       TestRun
VERSION:    v1alpha1

FIELD: envFrom <[]Object>

DESCRIPTION:
    <empty>
FIELDS:
  configMapRef  <Object>
    <no description>

  prefix        <string>
    <no description>

  secretRef     <Object>
    <no description>
```

I am curious to know if this has not yet been implemented or if this is a deliberate decision or if I am just missing something. 

This issue arises from a recent issue that I faced which I have [posted on Grafana k6-operator community forum](https://community.grafana.com/t/cannot-access-env-inside-initializer-pod-itself-and-testing-fails/109494/2)
 

3.2.0

### K6 YAML

```yaml
apiVersion: k6.io/v1alpha1
kind: TestRun
metadata:
  name: api-test
  namespace: podinfo
spec:
  arguments: -o experimental-prometheus-rw
  parallelism: 4
  script:
    configMap:
      name: api-cm-test
      file: api.js
  runner:
    env: 
      - name: K6_PROMETHEUS_RW_SERVER_URL
        value: http://localhost:9090/api/v1/write
      - name: K6_PROMETHEUS_RW_TREND_STATS
        value: p(95),p(99),min,max,avg,count,sum 
    envFrom: 
      - secretRef:
          name: api-secret
      - configMapRef: 
          name: api-data-cm
  initializer:
    env: 
      - name: "K6_INCLUDE_SYSTEM_ENV_VARS" 
        value: "1"
      - name: "RATE"
        value: '1'
      - name: "DURATION"
        value: '20s'
  # This does not work even though its in the spec
  # envFrom:
    #   - configMapRef:
    #       name: api-data-cm
    #   - secretRef:
    #       name: api-secret
```

GKE v1.27 

1. Create a ConfigMap with `RATE` and `DURATION` variables and a Kubernetes secret with values. 
2. Apply them.
3.  Use the following test as an example: 

```javascript
import http from 'k6/http';
import { check, fail } from 'k6';
import { sleep } from 'k6';

const url = 'https://url-to-test';

export let options = {
  scenarios: {
    constant_request_rate: {
      executor: 'constant-arrival-rate',
      rate: `${__ENV.RATE}`,
      timeUnit: '1s', // 1000 iterations per second, i.e. 1000 RPS
      duration: `${__ENV.DURATION}`,
      preAllocatedVUs: 5, // how large the initial pool of VUs would be
      maxVUs: 10, // if the preAllocatedVUs are not enough, we can initialize more
    },
  }
};

export default function() {
  let res = http.get(url, {
    headers: { 
      'Content-Type': 'application/json',
      'account-id': `${__ENV.ACCOUNTID}`,
      'api-key': `${__ENV.APIKEY}`,
      
    },
  });
  
  console.log(res.json()); 
}
```

### Expected behaviour

The tests should run correctly and should pick the values from the ConfigMap. 

### Actual behaviour

The initializer pod fails with the error that variables are undefined because it cannot pick up any values from the ConfigMap.

ghostx31:
As discussed, this is a very small PR adding the `envFrom` field for initializer and should fix #348 

This is my first time going through the k6-operator source code so if I missed anything (helm chart version bump or test cases in `initializer_test.go`) then let me know.

### DIFF
--- a/pkg/resources/jobs/initializer.go
+++ b/pkg/resources/jobs/initializer.go
@@ -123,6 +123,7 @@ func NewInitializerJob(k6 v1alpha1.TestRunI, argLine string) (*batchv1.Job, erro
 							Env:             env,
 							Resources:       k6.GetSpec().Initializer.Resources,
 							VolumeMounts:    volumeMounts,
+							EnvFrom:         k6.GetSpec().Initializer.EnvFrom,
 							Ports:           ports,
 						},
 					},
--- a/pkg/resources/jobs/initializer_test.go
+++ b/pkg/resources/jobs/initializer_test.go
@@ -65,7 +65,16 @@ func TestNewInitializerJob(t *testing.T) {
 								"sh", "-c",
 								"mkdir -p $(dirname /tmp/test.js.archived.tar) && k6 archive /test/test.js -O /tmp/test.js.archived.tar --out cloud 2> /tmp/k6logs && k6 inspect --execution-requirements /tmp/test.js.archived.tar 2> /tmp/k6logs ; ! cat /tmp/k6logs | grep 'level=error'",
 							},
-							Env:          []corev1.EnvVar{},
+							Env: []corev1.EnvVar{},
+							EnvFrom: []corev1.EnvFromSource{
+								{
+									ConfigMapRef: &corev1.ConfigMapEnvSource{
+										LocalObjectReference: corev1.LocalObjectReference{
+											Name: "env",
+										},
+									},
+								},
+							},
 							Resources:    corev1.ResourceRequirements{},
 							VolumeMounts: script.VolumeMount(),
 							Ports:        []corev1.ContainerPort{{ContainerPort: 6565}},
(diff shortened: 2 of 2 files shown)


ghostx31:
I've added the the test case for initializer and tested them with `make test` as well @yorugac

yorugac (approved):
@ghostx31 Thanks for the update :raised_hands:

Issue #309, opened 2023-10-12
Support `options.cloud`


olegbespalov:
# What?

In the https://github.com/grafana/k6/issues/1155 we're going to introduce the `options.cloud`, which should replace an existing `options.ext.loadimpact`

The k6-operator currently relies on the old version
https://github.com/grafana/k6-operator/blob/4a09026463a5f02f0c1c54cb873e321fd0b3f14a/pkg/cloud/types.go#L13-L18

We should implement support in both ways, with the priority on `options.cloud`

# Why?

Sunsetting of the loadimpact and in the future, the scripts will have only the `options.cloud`

PR #373, opened 2024-02-19
Adds support options.cloud

olegbespalov:
# What?

This change supports the `options.cloud` to the k6-operator. A logic mimics the one in the k6-core; the prior is for defining cloud options with a fallback to the load impact.

I haven't touched the README in that PR. I believe it could be updated along with updates to the k6-docs (once we are confident to do that).

### How did I test

TBH, I haven't done any testing apart from the running unit tests (`make test`). I also quickly searched for any existing tests that could be adjusted by introducing a new logic. I could add a simple test case for the unit testing the fallback logic of the new `TestName` and `ProjectID` functions if you think it's necessary or any other place that could be better for keeping those tests.

# Why?

The `options.cloud` refactor has been merged (https://github.com/grafana/k6/pull/3348) into k6-core, and the new way of defining options could appear soon.

Closes: https://github.com/grafana/k6-operator/issues/309


pkg/cloud/types.go
```diff
@@ -10,17 +10,44 @@ import (
...
+	External struct { // legacy way of defining the options.cloud
 		Loadimpact struct {
 			Name      string `json:"name"`
 			ProjectID int64  `json:"projectID"`
 		} `json:"loadimpact"`
 	} `json:"ext"`
+	Cloud struct { // actual way of defining the options.cloud
+		Name      string `json:"name"`
+		ProjectID int64  `json:"projectID"`
+	}
```
yorugac, line 22:
```suggestion
	} `json:"cloud"`
```
Right? This is supposed to be output from `k6 inspect`.

yorugac (changes requested):
Thanks for the PR, @olegbespalov.

This change looks like it should be backwards compatible so I assume it would work with any version of k6.

There is a question about exact format: please see my comment.

Regarding the unit test: given the long deprecation period ([years over here](https://github.com/grafana/k6/blob/b5a6febd56385326ea849bde25ba09ed6324c046/cloudapi/config.go#L198-L201)), yes, a small unit test(s) for the new logic would be nice.

yorugac (approved):
Thanks for the update, @olegbespalov. The change seems to be working as expected :+1: 

:information_source:  Until new release, the full impact of this change is noticeable only with `image: "grafana/k6:master"`.

Issue #375, opened 2024-02-23
Initializer fails on --execution-requirements


YuriiBudnyi:
### Brief summary

Hello Team. We only started to use K6 Operator in our environment. But we have an issue with Initializer, it fails on --execution-requirements if we have in test.js `const Users = __ENV.USERS.split(" ");` for example. The issue could be fixed if I add `--execution-requirements --include-system-env-vars` but I don't see possibility of changing Initializer command.

### k6-operator version or image

0.0.13

### Helm chart version (if applicable)

3.5.0

### TestRun / PrivateLoadZone YAML

```
apiVersion: k6.io/v1alpha1
kind: TestRun
metadata:
  name: test-k6
spec:
  cleanup: post
  initializer:
    envFrom:
      - secretRef:
          name: test-secrets
  parallelism: 1
  runner:
    envFrom:
      - secretRef:
          name: test-secrets
  script:
    configMap:
      file: test.js
      name: load-test
  separate: false

```

### Steps to reproduce the problem

Create test.js file with const Users = __ENV.USERS.split(" ");  
export USERS="TEST1 TEST2";
Run k6 inspect --execution-requirements test.js will result in error TypeError: Cannot read property 'split' of undefined or null
Run k6 inspect --execution-requirements --include-system-env-vars test.js result without errors

### Expected behaviour

Initializer not fails

### Actual behaviour

Initializer fails

yorugac:
Hi @YuriiBudnyi,
This looks like the same issue as here:
https://github.com/grafana/k6-operator/issues/207
And partially now described in the internal docs:
https://github.com/grafana/k6-operator/blob/main/docs/env-vars.md

As is described, there is a known discrepancy in how different k6 commands behave. It is not happening at the level of k6-operator. Adding `--include-system-env-vars` to initializer command means introducing potential security issues. IMHO, it'd be ideal to fix these issues at k6 level:
https://github.com/grafana/k6/issues/2744

Closer to your use case, would it work for you to move the logic around `Users` to [`setup`](https://grafana.com/docs/k6/latest/using-k6/test-lifecycle/) function instead of init context?

YuriiBudnyi:
Hello @yorugac,
Thank you very much for your response. For now, we will use the workaround described in internal docs:
`const Users = __ENV.USERS ? __ENV.USERS.split(" ") : [];`
Thanks for the help

yorugac:
[...]

Issue #207, opened 2023-03-22
Cannot pass env vars to k6 tests via k6-operator resource


danielfigueiredo:
### Brief summary

Hey team, I've been trying to get my k6 tests running via k8s using the `k6-operator`, but I'm unable to pass any environment variables.

I've looked in detail at the community forum and demo app, and whatever syntax is there does not seem to work for me.
The runnerSpec go file seems to forward the `env` vars declared within the `spec.runner`. However, my tests keep failing right when bootstrapping because the vars provided are unavailable.

I've tried several syntax combinations with different indentations, some of them failed to be parsed, and all others do not throw errors, but they don't seem to work either.

### k6-operator version or image

`make deploy` executed from commit sha `151472705d2e35d1c6d9158affed2ab127c305ba` (latest in the repo main branch at time of writing)

### K6 YAML
```yml
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6-distributed
spec:
  parallelism: 4
  script:
    configMap:
      name: my-test-file-dev
      file: my-test-file-dev.js
  runner:
    image: danielfigueiredoc/demo-k6-operator:latest
    env:
      - name: REMOTE_ENV
        value: staging
```

### Steps to reproduce the problem

My docker file is the same as the demo / recommended in k6-operator docs
```
FROM golang:1.19-alpine as builder
WORKDIR $GOPATH/src/go.k6.io/k6
ADD . .
RUN apk --no-cache add git
RUN CGO_ENABLED=0 go install go.k6.io/xk6/cmd/xk6@latest
# TODO - Want more extensions?! Provide additional `--with ...` lines to the following command:
RUN CGO_ENABLED=0 xk6 build \
    --output /tmp/k6

FROM alpine:3.16
RUN apk add --no-cache ca-certificates \
    && adduser -D -u 12345 -g 12345 k6
COPY --from=builder /tmp/k6 /usr/bin/k6

USER 12345
WORKDIR /home/k6

ENTRYPOINT ["k6"]
```

It is all very simple; tests run fine without env var assertions. But I added this check when my test bootstraps:
```js
import { fail } from 'k6';
//  ...

const env = __ENV.REMOTE_ENV;

if (!env) {
  fail('no REMOTE_ENV given');
}
```
When I run `kubectl apply -f ./path/to/k6-resource.yml` the `k6-distributed-initializer` job kicks in, I can see the pod being created and immediately failing with:
```
[...]

Issue #187, opened 2023-02-02
Support for linux/arm64/v8 devices such as (m1 mac)


anuragjanghala:
### Feature Description

Hey, please provide support for the arm64 devices, as after the recent changes, the new image with latest tag doesn't support m1 or m2 chips as it was doing in the previous release.

yorugac:
Hi @anuragjanghala,
I assume you were the one opening the forum question [here](https://community.k6.io/t/issues-with-k6-operator-setup-on-local-instance-of-k8s-cluster/5862), right?

There's actually a bit of misunderstanding: the operator has NOT supported arm64 at all, in any version. So the issue you encountered has a different cause.
The v0.0.9rc2 build has broken Docker manifests, due to an issue in underlying CI dependency. So the problem you encountered with v0.0.9rc2 is not related to ARM architecture. Please try to pull these images:
```
docker pull ghcr.io/grafana/operator:controller-v0.0.9rc2-buildx-fix
docker pull ghcr.io/grafana/operator:latest
```
These should have fixed manifests.

OTOH, I find it quite curious that previous versions of the operator have worked for you and other ARM users (since nobody had expressed an issue with it before). This implies that ARM-based devices that people use can actually run existing images out-of-the-box. You can check what image you've de facto pulled by running this command:
```
docker inspect ghcr.io/grafana/operator:latest | grep -A1 'Architecture'
```
I suspect that you'll see Linux amd64 in output, and your Docker setup must be doing some hidden magic with architectures. I'd appreciate feedback about that :slightly_smiling_face: It'd be good to understand better what exactly happens on ARM devices.

pasi-romo-idealo:
I have a related issue i.e. I tried to install the operator from the bundle:
https://raw.githubusercontent.com/grafana/k6-operator/main/bundle.yaml

but it fails to start on an arm64-node with following error:
`
k6-operator-controller-manager-5ddf587fdb-zk24j kube-rbac-proxy
k6-operator-controller-manager-5ddf587fdb-zk24j manager
[k6-operator-controller-manager-5ddf587fdb-zk24j manager] exec /manager: exec format error
`

The image, which is being used is this one:
ghcr.io/grafana/k6-operator:controller-v0.0.11rc2

If I pin the workload to run on an amd64-node it works without issues.

yorugac:
@pasi-romo-idealo, thanks for feedback! Can you clarify which arm64-node? I.e. is it a cloud cluster, etc.

[...]

PR #376, opened 2024-02-23
Include system env vars into Initializer job

ausias-armesto:
As part of my specific deployment, I needed to add the system environment variables into the Initializer Job.


yorugac:
Hi @ausias-armesto, thanks for the PR.
We've had and have issues on this topic:
https://github.com/grafana/k6-operator/issues/187
https://github.com/grafana/k6-operator/issues/375

Could you please describe details of your use case over there first?

On the subject of the change itself: ATM, I don't think it's a good idea to hard-code this option to all test runs in k6-operator. I'd suggest to look for a workaround first. If there is no workaround, it'd be good to understand why.

ausias-armesto:
> Hi @ausias-armesto, thanks for the PR. We've had and have issues on this topic: #187 #375
> 
> Could you please describe details of your use case over there first?
> 
> On the subject of the change itself: ATM, I don't think it's a good idea to hard-code this option to all test runs in k6-operator. I'd suggest to look for a workaround first. If there is no workaround, it'd be good to understand why.

Hi @yorugac ,

Thanks for the reply. I totally understand that the change may have impact on other test runs and I didn't consider that.

My repo is public and you can have a look: https://github.com/hoprnet/hoprd-test/tree/main/k6
I think that the reason was because it was needed  to built the [options variable](https://github.com/hoprnet/hoprd-test/blob/main/k6/src/k6/constant-traffic.test.ts#L40) which was created dynamically based on input environment variables. So I could use the same script and customise the environment variables named: ENVIRONMENT_NAME, WORKLOAD_NAME, SCENARIO_ITERATIONS, SCENARIO_DURATION, etc to the type of load testing execution that I need at any time. 
Perhaps there is a better alternative to implement that kind of requirements. 

This is the pipeline triggering the tests: https://github.com/hoprnet/hoprnet/blob/master/.github/workflows/load-tests.yaml

ausias-armesto:
This is definitely related to #375

yorugac:
In my prev comment: I meant issue https://github.com/grafana/k6-operator/issues/207 and not https://github.com/grafana/k6-operator/issues/187 :facepalm: 

Thanks for sharing details @ausias-armesto! AFAIS at the moment, it should be working fine assuming workloads are part of the [image](https://github.com/hoprnet/hoprd-test/blob/main/k6/assets/k6-test-run.yaml#L14) :thinking: What is the error you're getting when you try to run this?

ausias-armesto:
Hi @yorugac,

I've checked again with the current helm chart version `3.5.0` and it works fine without adding the system environment variables. I think that I tested with `3.3.0` and it failed. But it's good that it's not needed anymore.

Thanks.

ausias-armesto:
Hi @yorugac 
I've come up with the reason why it was working. I added a default value to the environment variables, and then those values were taken by the initialiser job but the issue still exists.
If the k6 script requires some environment variables to build the `options` then it requires that those environment variables are set with a default value at the code level

This snippet of code would fail.
```
const TEST_TYPE= __ENV.TEST_TYPE
var options
if(TEST_TYPE ="TEST_TYPE1") {
 options = ....
} else {
 options = ....
}
export options
```
But if the env variable has a default value like in the following snippet, then it works
```
const TEST_TYPE= __ENV.TEST_TYPE || "DEFAULT_TEST_TYPE"
var options
if(TEST_TYPE ="TEST_TYPE1") {
 options = ....
} else {
 options = ....
}
export options
```
Adding the parameter as implemented in the PR would solve the issue as well, but the developer needs to be aware of that restriction.

yorugac:
@ausias-armesto thanks for the update!

```js
const TEST_TYPE= __ENV.TEST_TYPE || "DEFAULT_TEST_TYPE"
```
This way of using env vars repeats very often actually. For instance, it's probably present in all examples in [quickpizza](https://github.com/grafana/quickpizza). Like a "k6 good practice" thing.

Perhaps, it could be pointed out more, esp. in the [main docs](https://grafana.com/docs/k6/latest/using-k6/environment-variables/) :thinking: I can also add an additional emphasis on it in [k6-operator docs for env vars](https://github.com/grafana/k6-operator/blob/main/docs/env-vars.md), but it sounds like it is a proper workaround without having to pass system env vars to the jobs.

Issue #296, opened 2023-09-26
Add support for separate initContainers in initializer spec


nveces:
### Brief summary

I have a K6 yaml with the "initializer > initContainers" and "runner > initContainers":
```
initializer:
    initContainers:
      - name: test-init-01-clone
        image: busybox:1.28
        command: ['sh', '-c', 'echo "InitiContainer Waiting for task...."; sleep 10; echo "Task is done!!!"']
.../...
  runner:
    initContainers:
      - name: test-runner-01
        image: busybox:1.28
         command: ['sh', '-c', 'echo "Waiting for task...."; sleep 10; echo "Task is done!!!"']    
```

But, the "k6-loadtest-01-initializer-<xxxx>" pod shows the "test-runner-01" init-container.

I'm confused, I thought that the initializer pod would use the initializer and the runner pods the "runner > initContainer".

I want to clone a git repo at the begins of the initializer pod, this repo contains the Load Tests scripts and libraries and leaves the files on the volumepersistentclaim (PVC).

### k6-operator version or image

`ghcr.io/grafana/k6-operator:controller-v0.0.11rc1`

### K6 YAML

```
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6-loadtest-01
  namespace: k6-operator-system
  labels:
    app: k6-sso-loadtest-01
spec:

  initializer:
    initContainers:
      - name: test-init-01-clone
        image: busybox:1.28
        command: ['sh', '-c', 'echo "InitiContainer Waiting for task...."; sleep 10; echo "Task is done!!!"']
        volumes:
          - name: clone-repo
            gitRepo:
              directory: /test/project
              repositoy: https://anexamplepapthofgitlab.org/k6-loadtest.git
              revision: main
        volumeMounts:
          - name: clone-repo
            mountPath: /test
            readOnly: false
  
  parallelism: 1
  script:
    volumeClaim:
      name: k6-projects-pvc
      file:  project/loadtest/10-test.js
  runner:
    initContainers:
      - name: test-runner-01
        image: busybox:1.28
        command: ['sh', '-c', 'echo "Waiting for task...."; sleep 10; echo "Task is done!!!"']
```

### Steps to reproduce the problem

1/ Login in OpenShift/KUbernetes cluster
2/ Execute: oc apply -f k6-test-load.yaml
3/ go to the k6-operator-system namespace
[...]

PR #379, opened 2024-02-29
feat: differentiate between Runner and Initializer initContainers

FloGro3:
Fixes [296](https://github.com/grafana/k6-operator/issues/296)

Enables you to have separate initContainers for the initializer and runner pod. 

Was tested successfully in own K8 environment (only the initializer pod was configured to have initContainers):
image

Config looks like this:
```yaml
initializer:
    ...
    {{- if .Values.executor.script.volumeClaim }}
    initContainers:
      - image: {{ tpl $.Values.docker.registry $ }}/{{ tpl $.Values.image.name $ }}:{{ tpl $.Values.image.tag $ }}
        {{- if .Values.image.command }}
        command: [ "sh", "-c", {{ .Values.image.command }} ]
        {{- else }}
          {{- with (index .Values.mount.pvcs 0) }}
        command: [ "sh", "-c", "cp -r /home/* {{ .path }};" ]
          {{- end }}
        {{- end }}
    {{- end }}
```

### DIFF
--- a/config/samples/k6_v1alpha1_k6_with_initContainers.yaml
+++ b/config/samples/k6_v1alpha1_k6_with_initContainers.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: k6.io/v1alpha1
+kind: K6
+metadata:
+  name: k6-sample
+spec:
+  parallelism: 4
+  script:
+    configMap:
+      name: k6-test
+      file: test.js
+  initializer: 
+    initContainers:
+      - image: busybox:latest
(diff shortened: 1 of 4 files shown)


yorugac (changes requested):
@FloGro3, thanks for working on this! Could you please add an example for this case to `config/samples`?

FloGro3:
@yorugac Sure, I've added a small example.

yorugac (approved):
Apologies for such a delay, @FloGro3!
LGTM, merging in now. Thanks again for working on this :raised_hands:

PR #289, opened 2023-09-18
Adding k6-browser support to runner image

javaducky:
Add support for distributed k6-browser scripts.

PR #380, opened 2024-03-01
Add arm64 support

audacioustux:
solves https://github.com/grafana/k6-operator/issues/187

Pls make any changes if required.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -33,7 +33,7 @@ jobs:
       - name: "Build:buildx"
         uses: docker/setup-buildx-action@v3
         with:
-          version: v0.12.1 # Buildx version
+          version: v0.9.1 # Buildx version
       - name: "Build:login"
         uses: docker/login-action@v3
         with:
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
     inputs:
       image_tag:
-        description: 'Image tag name.'
+        description: "Image tag name."
         required: true
   release:
     types:
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: "Build:checkout"
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
       - name: "Set image tag name"
         run: |
           if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
@@ -28,38 +28,43 @@ jobs:
       - name: "Check image tag name"
         run: |
           echo "IMAGETAG=${{env.IMAGETAG}}"
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
       - name: "Build:buildx"
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
         with:
-          version: v0.9.1 # Buildx version
+          version: v0.12.1 # Buildx version
       - name: "Build:login"
-        uses: docker/login-action@v1
+        uses: docker/login-action@v3
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: "Build:dockerimage"
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v5
         with:
           context: .
           push: true
           file: Dockerfile.controller
           build-args: |
             GO_BUILDER_IMG=golang:1.19
+          platforms: linux/amd64,linux/arm64
           tags: ghcr.io/grafana/k6-operator:latest,ghcr.io/grafana/k6-operator:controller-${{env.IMAGETAG}}
       - name: "Build:dockerimage"
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v5
         with:
           context: .
           push: true
           file: Dockerfile.runner
+          platforms: linux/amd64,linux/arm64
           tags: ghcr.io/grafana/k6-operator:latest-runner,ghcr.io/grafana/k6-operator:runner-${{env.IMAGETAG}}
       - name: "Build:dockerimage"
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v5
         with:
           context: .
           push: true
           file: Dockerfile.starter
+          platforms: linux/amd64,linux/arm64
           tags: ghcr.io/grafana/k6-operator:latest-starter,ghcr.io/grafana/k6-operator:starter-${{env.IMAGETAG}}
 
   bundle:
@@ -70,7 +75,7 @@ jobs:
       pull-requests: write
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
       - name: "Set image tag name"
         run: |
           if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
--- a/Dockerfile.starter
+++ b/Dockerfile.starter
@@ -1,9 +1,3 @@
-FROM radial/busyboxplus:curl
-
-# as defined by distroless for nonroot
-USER 65532:65532
-COPY --from=redboxoss/scuttle:latest /scuttle /bin/scuttle
-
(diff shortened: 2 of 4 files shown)


.github/workflows/unit-test.yaml
```diff
@@ -7,17 +7,17 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.19.x]
-        k8s_version: [1.19.2, 1.24.1, 1.27.1]
+        go-version: [1.19]
+        k8s_version: [1.24.1, 1.27.1, 1.28.3]
```
yorugac, line 11: Please remove these changes.

.github/workflows/e2e-test.yaml
```diff
@@ -9,23 +9,23 @@ jobs:
       - name: "Build:checkout"
         uses: actions/checkout@v2
       - name: "Build:buildx"
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
         with:
-          version: v0.9.1 # Buildx version
```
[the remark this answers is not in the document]
audacioustux: I see... the version bump worked for the push.yml workflow, so didn't check if it works for the e2e, my bad. Will check whether the e2e works completely fine.

audacioustux:
@yorugac thanks for the feedback, I'll send individual PRs as you mentioned.
Yes, have checked the ARM Images, and it works https://github.com/audacioustux/k6-operator/pkgs/container/k6-operator

yorugac:
Hi @audacioustux, it's been awhile; hope you're well :slightly_smiling_face: 
How is it going? Would you be able to continue with this PR?

audacioustux:
Hi @yorugac, sorry got busy with other stuffs. https://github.com/audacioustux/k6-operator I could make the ARM64 support work here, but that was a bit too involved than I had anticipated. + The fact that there's hard-coded image uri in the code made it a bit difficult to make a well-tested PR. 
Also, I couldn't make the browser tests work https://github.com/grafana/xk6-browser/issues/1239 (I think it's not related to Arm though) and couldn't manage the time to look into it either.

I think the first thing that needs to be done, is to fix that hardcoded container image URI issue (for example: https://github.com/grafana/k6-operator/blob/c2e81d8c83297561e4539a2377dbd90ad9dab887/pkg/resources/jobs/stopper.go#L19) 

For the ARM64 support, I can hopefully send the PR in next weekend (30/31 March), sorry for keep you waiting.

yorugac:
Hi @audacioustux, thanks for quick response!

Yes, the browser tests are a "known" issue: #289 AFAIR, some cases have worked and some didn't. Investigating this wasn't prioritized yet but good thing you've opened the issue in xk6-browser repo as well :+1: 

> I think the first thing that needs to be done, is to fix that hardcoded container image URI issue

:thinking: you meant for testing? To test custom images, you can pass them as part of `TestRun` CR definition, i.e. `starter.image` in this example.

> For the ARM64 support, I can hopefully send the PR in next weekend (30/31 March)

That would be amazing; thank you.

Btw, re-reading this PR: I've asked to have CI updates as separate PRs -- would you like me to take those, to reduce the "change request"?

audacioustux:
> you meant for testing? To test custom images, you can pass them as part of TestRun CR definition, i.e. starter.image in this example.

Umm, as far I can remember, even though starter and runner image was declared in TestRun manifest, it was spawning grafana/k6-operator. May have to re-check that actually...

> Btw, re-reading this PR: I've asked to have CI updates as separate PRs -- would you like me to take those, to reduce the "change request"?

Yes please.

audacioustux:
@yorugac Hi, pls let me know if the push works correctly and if anything needs to be changed.

Dockerfile.starter
```diff
@@ -1,5 +1,3 @@
-FROM radial/busyboxplus:curl
+FROM curlimages/curl:latest
 
-# as defined by distroless for nonroot
-USER 65532:65532
-COPY --from=redboxoss/scuttle:latest /scuttle /bin/scuttle
+COPY --from=kvij/scuttle:latest /scuttle /bin/scuttle
```
yorugac, line 3: `nonroot` user definitely shouldn't be removed here. 
On that note, are these two Docker changes necessary for ARM support? If they aren't, again, I'd suggest moving them out into a separate PR.

yorugac (changes requested):
Hi @audacioustux, Thanks for the update!

On CI changes in `push.yaml`: I've gone through these ones by now and they seem to be mostly OK.
`buildx` question remains: it seems like the issue we encountered was around ~ `v0.10` so maybe `v0.12` will be fine. If this version change is not necessary for ARM, I'd prefer to see it in another PR.

Secondly, on Dockerfiles: there is a change request there.

> let me know if the image build and push works correctly

To test this workflow (without further changes), I'll need to merge this PR first :joy: In general, that's why I request as minimal version of changes as possible: it can be pretty time-consuming to debug CI issues so it's best when they're small and atomic.

Dockerfile.starter
audacioustux: yes, those are required for ARM64 support, as the current ones are old/unmaintained and lacks arm64 support. is the USER really necessary for curlimages/curl:latest(https://github.com/curl/curl-container)?

audacioustux:
> On CI changes in `push.yaml`: I've gone through these ones by now and they seem to be mostly OK. `buildx` question remains: it seems like the issue we encountered was around ~ `v0.10` so maybe `v0.12` will be fine. If this version change is not necessary for ARM, I'd prefer to see it in another PR.

umm which issue exactly? I updated the version just to be on safe side and hoped it's most likely backward compatible. I wasn't sure if the "platform" value is supported in the previous version, so tried to save one trial and error.

> To test this workflow (without further changes), I'll need to merge this PR first 😂 In general, that's why I request as minimal version of changes as possible: it can be pretty time-consuming to debug CI issues so it's best when they're small and atomic.

I see. that's actually the issue I was referring to earlier. I think the current situation of the repo is a bit hard to rollout new changes or for community contributions. I actually had plan for adding Devcontainers support, that I think should significantly make things easier and streamlined - we may discuss about that later at some point.

Dockerfile.starter
yorugac: >  those are required for ARM64 support, as the current ones are old/unmaintained and lacks arm64 support
OK, thanks for clarification.

> is the USER really necessary for curlimages/curl:latest(https://github.com/curl/curl-container)?
Yes. It was added deliberately. Currently it is present in both controller and starter; please keep those.

yorugac:
On `buildx`: yes, I meant that issue. Basically Docker manifests got broken with newer version. I've been checking the GH changelogs: it seems like it's _still_ an issue even now... There might be a way to make a fix without pinning a version but for now let's not change that line with `buildx` version. I think ARM build should work anyway, right?

.github/workflows/push.yaml
```diff
@@ -28,38 +28,43 @@ jobs:
...
         run: |
           echo "IMAGETAG=${{env.IMAGETAG}}"
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
       - name: "Build:buildx"
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
         with:
-          version: v0.9.1 # Buildx version
+          version: v0.12.1 # Buildx version
```
yorugac, line 36: IOW, without this bump ^

Dockerfile.starter
audacioustux: >  Currently it is present in both controller and starter; please keep those.

added USER 65532:65532 now

.github/workflows/e2e-test.yaml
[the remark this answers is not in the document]
audacioustux: buildx version reverted back to 0.9.1 now

audacioustux:
https://github.com/audacioustux/k6-operator/actions/runs/8500731873/job/23282902694 the push.yml worked with the buildx version pinned to 0.9.1, kept the `USER 65532:65532` in starter dockerfile... tested in k8s and working

yorugac (approved):
Great, thanks @audacioustux! :raised_hands:

Israphel:
which version has ARM support?

PR #409, opened 2024-06-04
Set manager as first container in deployment

TheoBrigitte:
This sets `manager` container as default container in the `Deployment`.

This allow for `manager` container to be the default container shown when using `kubectl logs` otherwise one need to add `--container=manager` in order to get logs from the manager, since `kube-rbac-proxy` is the current default.


yorugac (changes requested):
Hi @TheoBrigitte, thanks for the PR. I've wondered if anyone needs that change with containers but it never came up as an issue before. Technically, it still didn't since this is a PR, not an issue :slightly_smiling_face: 

I think overall this change shouldn't be breaking but if there are concerns from anyone, please let me know.

On a PR itself: this change needs to be done in Helm chart as well, to remain consistency.

yorugac (approved):
Thanks for the update @TheoBrigitte! LGTM :raised_hands: 

I'll merge this change next week: if there are any concerns about it within a community, please comment here.

AFAIK, the only possible impact is if anyone is referencing containers by index somewhere, like `containers[1]`. That can happen in kustomizations, for instance.

Issue #279, opened 2023-08-24
Support ROX(Read Only Many) PVC as file source.


mugioka:
### Feature Description
ROX(Read Only Many) PVC should be supported as they allow load testing distributed across multiple nodes.
Now, k6-operator does not support ROX PVC as file source.
To do so we will need to support ReadOnly parameter in Volumes/VolumeMounts.
https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/readonlymany-disks
> In your Pod specification, you must specify readOnly: true in the volumeMounts section and the volumes section, such as in the following example:

### Suggested Solution (optional)

```yaml
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: k6
spec:
  script:
    volumeClaim:
      name: scenario-resources
      readOnly: true
```

yorugac:
Hi @mugioka, thanks for the issue!

I can't help but wonder if this (`readOnly: true`) shouldn't be a hard-coded default instead of adding a new field to CRD, esp. now that there are volumes in init containers :thinking:

mugioka:
@yorugac 
Considering init containers, I believe a new field should be added to the CRD to avoid making destructive changes to the k6-operator, since they are likely to modify the mounted volumes.

yorugac:
Closed via #415

PR #415, opened 2024-06-12
Add ReadOnly option to volumeMount in Pods

irumaru:
Added ReadOnly to the volumeMount option in Pods to support mounting of ReadOnlyMany(ROX) PVC.
https://github.com/grafana/k6-operator/issues/279

The default is ReadOnly: false.
The reason was to avoid destructive changes, considering the possibility that some users may already be writing to the volume in the init container.

Here are the results of a simple test.
[files.zip](https://github.com/user-attachments/files/15799552/files.zip)

Test Contents
- kubectl apply -f pvc.yaml to create pods to add pods and scenarios
- kubectl exec -it vp -- /bin/bash to add main.js to /mnt/vp in pods
- Create TestRun with kubectl apply -f main.yaml

result
describe-log.txt and get-log.txt are the execution results.
You can confirm that ReadOnly for k6-test-volume is true.

### DIFF
--- a/pkg/types/script.go
+++ b/pkg/types/script.go
@@ -10,6 +10,7 @@ import (
 // Internal type created to support Spec.script options
 type Script struct {
 	Name     string // Name of ConfigMap or VolumeClaim or "LocalFile"
+	ReadOnly bool
 	Filename string
 	Path     string
 	Type     string // ConfigMap | VolumeClaim | LocalFile
@@ -29,6 +30,7 @@ func (s *Script) Volume() []corev1.Volume {
 				VolumeSource: corev1.VolumeSource{
 					PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
 						ClaimName: s.Name,
+						ReadOnly:  s.ReadOnly,
 					},
 				},
 			},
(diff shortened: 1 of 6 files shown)


pkg/types/script.go
```diff
@@ -10,6 +10,7 @@ import (
 // Internal type created to support Spec.script options
 type Script struct {
 	Name     string // Name of ConfigMap or VolumeClaim or "LocalFile"
+	ReadOnly bool
```
yorugac, line 13:
```suggestion
	ReadOnly bool // VolumeClaim only
```

yorugac (changes requested):
@irumaru, sorry for the delay!
Thanks for working on this :raised_hands: PR LGTM: there's just a minor suggestion to improve comments. And one more thing: could you please create a folder in `config/samples` with code files from your archive?

irumaru:
@yorugac , Thanks for the review.

We have made the following two corrections.
- Added `config/samples/k6_v1alpha1_k6_with_readOnlyVolumeClaim.yaml`.
- Added comment to `pkg/types/script.go` as suggested.

Please review.

yorugac (approved):
Thank you for the update @irumaru! :raised_hands:

Issue #171, opened 2022-12-05
Starter doesn’t work in an IPv6 cluster


yann-soubeyrand:
### Brief summary

In an IPv6 cluster, the starter pod logs this error:
```
curl: (3) IPv6 numerical address used in URL without brackets
```

### k6-operator version or image

latest

### K6 YAML

```
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  name: load-test
  namespace: load-testing
spec:
  parallelism: 100
  script:
    configMap:
      name: load-test
      file: load-test.js
  separate: true
  runner:
    resources:
      requests:
        cpu: 1
        memory: 4G
```

### Steps to reproduce the problem

Create an IPv6 cluster on AWS EKS, deploy the operator inside and start a test.

### Expected behaviour

The test run as expected.

### Actual behaviour

The test doesn’t start because the operator doesn’t seem to handle IPv6.

yorugac:
Looks like it's a rather well-known issue for [IPv6 in general](https://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers) and specifically for [curl](https://stackoverflow.com/questions/24302447/ipv6-curl-post-request#24302866).

At first glance, seems like a quick fix with adding brackets to curl command. But curl doesn't seem to support brackets with IPv4 so we'll need to check first if it's IPv4 or IPv6; maybe before Job creation, in https://github.com/grafana/k6-operator/blob/5cec748534424741fbea0e1875218b56bf5af670/controllers/k6_start.go#L71-L80

Either way, thanks for opening this, @yann-soubeyrand!

pasi-romo-idealo:
I have the same issue as described earlier i.e. I try to start a test on an ipv6-only EKS-cluster, but the k6-sample-starter-pod fails to start due to following error:

curl: (3) IPv6 numerical address used in URL without brackets

For now I was able to go around this issue by creating a custom-image by using following line:

`
parts = append(parts, fmt.Sprintf("curl -g --retry 3 -X PATCH -H 'Content-Type: application/json' http://[%s]:6565/v1/status -d '%s'", hostname, req))
`
in NewStartContainer and NewStopContainer-function (please note that I needed to provide also this -g globbing option)

If there is away to do this globally i.e. supporting both ipv4 and ipv6, I would be more than happy to try that out :)

kahirokunn:
Let’s go ahead 🙆

zzhao2010:
[...]

PR #437, opened 2024-08-07
Fix IPv6 compatibility issue with starter pod

zzhao2010:
**Summary**
Bug Fixed: [Starter doesn't work in an IPv6 cluster](https://github.com/grafana/k6-operator/issues/171)

This pull request addresses the IPv6 compatibility issue in the k6-operator as outlined in the referenced issue. Specifically, the changes ensure that the k6-curl container can correctly handle IPv6 addresses.

**Changes Made**
Introduced the net package to handle IPv6 addresses.
Hostname Wrapping: Utilized net.JoinHostPort to wrap the hostname in square brackets if it is a literal IPv6 address, ensuring proper formatting.

**Testing**
Verified that the k6-curl container now supports both IPv6 and IPv4 addresses.
Conducted tests on local k8s cluster via KIND to confirm the functionality of the k6 operator with the updated IPv6 support.

Snapshot of starter pod that deals with IPv6 address correctly:
image
Snapshot of starter pod that still works with IPv4 address:
image

**Impact**
These changes enhance the compatibility and robustness of the k6-operator, allowing it to function correctly in environments utilizing IPv6 addresses.

### DIFF
--- a/e2e/ipv6/kind-ipv6.yaml
+++ b/e2e/ipv6/kind-ipv6.yaml
@@ -1,0 +1,5 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  ipFamily: ipv6
+  apiServerAddress: 127.0.0.1
--- a/pkg/resources/containers/curl_start.go
+++ b/pkg/resources/containers/curl_start.go
@@ -3,6 +3,7 @@ package containers
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/grafana/k6-operator/pkg/types"
@@ -26,7 +27,7 @@ func NewStartContainer(hostnames []string, image string, imagePullPolicy corev1.
 
 	var parts []string
 	for _, hostname := range hostnames {
-		parts = append(parts, fmt.Sprintf("curl --retry 3 -X PATCH -H 'Content-Type: application/json' http://%s:6565/v1/status -d '%s'", hostname, req))
+		parts = append(parts, fmt.Sprintf("curl --retry 3 -X PATCH -H 'Content-Type: application/json' http://%s/v1/status -d '%s'", net.JoinHostPort(hostname, "6565"), req))
 	}
 
 	return corev1.Container{
--- a/pkg/resources/containers/curl_stop.go
+++ b/pkg/resources/containers/curl_stop.go
@@ -3,6 +3,7 @@ package containers
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/grafana/k6-operator/pkg/types"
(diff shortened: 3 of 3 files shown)


yorugac (changes requested):
Hi @zzhao2010, thank you for the PR!
I like this solution: it's super-concise :smile: 

It should be tested a bit more though. I'll be able to look into how to test it internally only in a couple of weeks. Meanwhile, could you please add reproducible examples with kind cluster? It appears they have it here:
https://kind.sigs.k8s.io/docs/user/configuration/#ip-family
There is separately `ipv6` and `dual` configuration: I think we should be checking both separately.

If you could add a folder `e2e/ipv6` with relevant files for such a test, that'd be great.

frittentheke:
While I highly appreciate IPv6 to be working ... and this PR is nothing more than a bugix. But there also is https://github.com/grafana/k6-operator/issues/87. 

If you, @yorugac, agree to move to a native k6 client / HTTP call by the Operator to kick off them K6 runners that change can also "fix" the IPv6 endpoint issue.

LCaparelli:
> While I highly appreciate IPv6 to be working ... and this PR is nothing more than a bugix. But there also is https://github.com/grafana/k6-operator/issues/87.

I don't think one solution excludes the other. The fix with brackets is here, and would be greatly appreciated. For now we're using a fork.

I'd suggest attacking both fronts: this fix as a quick win and #87 as a more robust, long-term approach. :-)

zzhao2010:
> PS Signing CLA is a hard requirement for any PR: please sign it 🙂

Hi @yorugac,
I did sign CLA. But not sure why it still shows as "not signed yet".
image

zzhao2010:
Hi @yorugac 
> could you please add reproducible examples with kind cluster? It appears they have it here:
https://kind.sigs.k8s.io/docs/user/configuration/#ip-family
I assume you mentioned the same in your description.

That is correct. I created the ipv6 local cluster via KIND using the exactly same config file in this url. I did test it using both ipv4 and ipv6 cluster.

> If you could add a folder e2e/ipv6 with relevant files for such a test, that'd be great.

You needed me to add the kind config file in the e2e/ipv6 folder? For the test script and crd file, the test.js and test.yaml that are already in e2e folder work.

> PS Signing CLA is a hard requirement for any PR: please sign it 🙂

Signed.

yorugac (approved):
Hi @zzhao2010, thank you for the update! LGTM.

Noting for posterity's sake. Depending on the setup, pure `ipFamily: ipv6` might fail on DNS resolution. `ipFamily: dual` seems more reliable in this context.
Potentially related: https://github.com/kubernetes-sigs/kind/issues/3114

yorugac:
> I'd suggest attacking both fronts: this fix as a quick win and https://github.com/grafana/k6-operator/issues/87 as a more robust, long-term approach. :-)

Yes, I think the same. Removal of starter won't happen fast, but this fix is small and a quick win :raised_hands:

zzhao2010:
Hi @yorugac Can I ask when are you planning to build a image that includes this fix? The latest image was built 2 months ago.

yorugac:
@zzhao2010, yes, there was a bit of a skew with the releases last time: 2 releases were done with the short window. The next one will be after k6 v0.54 so ~ next week. (k6 will be released on Tuesday)
And yes, in case anyone is wondering, we do intend to improve this workflow so that releases of k6-operator are less dependent on k6 ([issue](https://github.com/grafana/k6-operator/issues/452)).

liyuntian:
> skew
the Dockerfile.runner is 
e0ddfd21985d06c6c9aca9ce0e144d3

If you have modified the k6-operator code, what does the Dockerfile need to be changed ,in order to include the modified code in the image? What is your Dockerfile? thank you

liyuntian:
> handle IPv6 addresses.

or can you share your "handle IPv6 addresses " image

PR #438, opened 2024-08-12
Increase readiness and liveness probe timeouts, introduce an initial delay

### LINKED ISSUE #306: Failures of default liveness / readiness probes
Several times we have received reports of failing default liveness and readiness probes which leads to runners not reaching ready state:
```
  Normal   Created    3m2s                   kubelet            Created container k6
  Normal   Started    3m2s                   kubelet            Started container k6
  Warning  Unhealthy  2m34s (x7 over 3m1s)   kubelet            Readiness probe failed: Get "http://10.1.0.75:6565/v1/status": dial tcp 10.1.0.75:6565: connect: connection refused
  Warning  Unhealthy  2m34s (x3 over 2m54s)  kubelet            Liveness probe failed: Get "http://10.1.0.75:6565/v1/status": dial tcp 10.1.0.75:6565: connect: connection refused
  Normal   Killing    2m34s                  kubelet            Stopping container k6
```
Such error does not allow to start the test and the test is stuck (at the moment, but this part will likely change as result of https://github.com/grafana/k6-operator/issues/260).

Default probes are the same defaults that are set by Kubernetes, i.e. k6-operator does not add anything by itself. k6 HTTP starts pretty quickly for most tests and it should be well reachable with default settings normally. The most confounding part is that this failures seemed to have been reported while using simple and basic setups of Kubernetes, even local clusters.

So far it is not clear how to repeat this case. If someone knows a trick how to repeat this, please share in this issue.

cc @dgzlopes

frittentheke:
Currently the default timeout of 1 second and no initial delay is applied to the probes of the runner pods. Depending on the startup time this can cause random Pod errors causing a whole TestRun to fail.

At some point it might also make sense to introduce a startupProbe to cover the longer initial startup time a K6 instance (Pod) might need instead of every increasing the runtime liveness and readiness checks.

Fixes #306


frittentheke:
@yorugac Since you opened the issue (https://github.com/grafana/k6-operator/issues/306) PTAL at this proposed fix / mitigation.

As commented, I believe a startup probe would make some sense and also distinguishing between liveness and readiness checks (which is a different thing).

vipul-mykaarma:
any eta when this will go live ? ie increase the timeout of probes

PR #443, opened 2024-08-21
Golang and dependencies update

yorugac:
Bumps:
- Golang to 1.22
- controller-gen to 0.16.1
- controller-runtime to 0.19.0

Additionally, some cleaning up of Makefile.


frittentheke:
@yorugac maybe it makes sense to just adopt Dependabot like K6 uses itself (https://github.com/grafana/k6/blob/master/.github/dependabot.yml) ?

yorugac:
:wave: I've missed the question here :sweat_smile: Dependabot is set up for this repository as well, for security vulnerabilities. There was no need to have a `.yml` config so far.

Issue #193, opened 2023-02-07
Cloud output: make reading logs by initializer a more "secure" operation


yorugac:
Envoy is not usable with cloud output when logging is enabled. See https://github.com/grafana/k6-operator/issues/179. It’d be good to figure out some alternative in order not to depend on logs in the initializer so much.

Context. Previously, adding k6 dependency to the operator to calculate MaxVUs and Duration was deemed as too heavy to add directly - that's how initializer job appeared. Check if that changed in recent refactoring of k6.

yorugac:
Additionally, k6 Operator is not usable in some specific cases of k6 scripts. It was reported in #207 and see explanation [here](https://github.com/grafana/k6-operator/issues/207#issuecomment-1482564043).

k6 docs about `k6 archive`, `k6 inspect` and `k6 cloud` behaving differently than `k6 run` around env vars:
https://k6.io/docs/using-k6/environment-variables/

It's a bit of mess of inconsistent behaviour. It'd be good to have a generalized solution to this, in order not to stumble over k6 exceptions.

yorugac:
The same problem can manifest even without cloud output but in any scenario when there is another process writing to the output of the initializer Pod. Which sounds a bit strange at first glance but it does happen, case in study:
https://community.grafana.com/t/k6-operator-error-not-spawning-starter-and-basic-resources/110778/19

Renaming the issue to reflect new reality.

yorugac:
After some additional thinking and feedback :slightly_smiling_face: 

Initializer pod is currently used for two reasons:
1. cloud output tests require duration and max VUs values in order to be started: those are retrieved with `k6 inspect` command.
2. prior validation of a script. k6 Operator test runs can have lots of runners and before scheduling them, it makes sense to at least confirm that the script is valid.

If there is a way to implement this without running a pod but by calling k6 Go library, we could remove initializer pod completely! That would simplify running tests for users (because of initializer-specific errors mentioned above), simplify TestRun CRD definition and simplify logic of TestRun controller.
Now to figure out if it can be done with k6 Go code...

yorugac:
Another important reason to solve this: regular issues with configuring env vars for k6 scripts. The error appears because of k6 quirks and inconsistencies between different commands, and people regularly stumble on this. Some sample issues:
- https://github.com/grafana/k6-operator/issues/375
- https://github.com/grafana/k6-operator/pull/376
- https://github.com/grafana/k6-operator/issues/207
- forum questions

[...]

PR #450, opened 2024-08-22
Rework Initializer container and handling its verdict

### LINKED ISSUE #435: Issues within initalizer error handling if script is incorrect
### Brief summary

We realized, that our TestRuns get stuck without any information/logs printed out if the script itself is incorrect.

# Cause

We already had a deeper look and realized that this is likely related to the Log message parsing of the `k6 inspect` executed within the initalizer here: https://github.com/grafana/k6-operator/blob/f75facb321d3c8ca55bbd9ba2f1895173d10bbc7/pkg/resources/jobs/initializer.go#L79

When we execute `k6 inspect` manually inside the container, we get the following error:
```sh
/ # k6 inspect --execution-requirements  /test/<REDACTED>.js
ERRO[0009] could not initialize '/test/<REDACTED>.js': could not load JS test 'file:///test/<REDACTED>.js': unknown executor type 'burst' 
```

The log parsing mentioned above, basically does a `| grep 'level=error`, which does not work for the error message we are facing since the log format seems to be different.

Might be related to:  https://github.com/grafana/k6-docs/issues/877

ghcr.io/grafana/k6-operator:controller-v0.0.14

_No response_

Since we built a custom k6 image to include the script to be used as localfile, it would need some additional effort to make this available.

However, IMO this is not really related to a specific TestRun.

k6 version: k6 v0.51.0 (go1.22.4, linux/amd64)

- Deploy any TestRun resource with a broken JS script

### Expected behaviour

- Initalizer Pods fails or at least logs the issues which are encountered during `k6 inspect`
- TestRun resource is in Failed state

### Actual behaviour

- Initalizer Pod runs but does not print out anything
- The k6-operator-controller-manager logs errors (see below) about unparsable JSON without any information
- TestRun resource stuck in `initilization` phase

<details>
  <summary>Logs of k6-operator</summary>

```
2024-07-30T09:11:00Z    ERROR   controllers.TestRun     unable to marshal: ``   {"namespace": "loadtesting", "name": "<REDACTED>", "reconcileID": "<REDACTED>", "error": "unexpected end of JSON input"}
github.com/grafana/k6-operator/controllers.inspectTestRun
        /workspace/controllers/common.go:105
github.com/grafana/k6-operator/controllers.RunValidations
        /workspace/controllers/k6_initialize.go:55
github.com/grafana/k6-operator/controllers.(*TestRunReconciler).reconcile
        /workspace/controllers/testrun_controller.go:137
github.com/grafana/k6-operator/controllers.(*TestRunReconciler).Reconcile
        /workspace/controllers/testrun_controller.go:80
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:119
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:227
2024-07-30T09:11:00Z    ERROR   Reconciler error        {"controller": "testrun", "controllerGroup": "k6.io", "controllerKind": "TestRun", "TestRun": {"name":"<REDACTED>","namespace":"loadtesting"}, "namespace": "loadtesting", "name": "<REDACTED>", "reconcileID": "<REDACTED>", "error": "unexpected end of JSON input"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:329
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.2/pkg/internal/controller/controller.go:227
```
</details>

frittentheke:
Use exit codes (of k6 archive + inspect) as sole indication for initialization success

1) Introduce an `EmptyDir` volume for temporary data
This removes the need for any mkdir and avoids writing to the container filesystem,
allowing for e.g. `readOnlyRootFilesystem` security contexts ([1]).

2) Remove all output redirection and grepping for error indicating strings in favor of
using the exit codes from the commands to indicate success or failure reestablishing the 
clear contract of the `inspect` command to judge the provided config.

This approach is much cleaner in regards to any changes to the output of K6 and the vast
space of error there is.

`k6 inspect --execution-requirements` ([2,3]) while also printing warnings or the JSON
should clearly indicate via `err` if there are issues with the provided tests, which 
are then converted to a non-zero RC ([4]).

3) Set TerminationMessagePolicy on Initializer to use logs as fallback

While not explicitly used, with the removal of the log fetching from the  
Initializer (Job->Pod->Container) any potential source for verbose information
about the error might be lost. The Kubernetes approach to provide more details
than the exit code about the reason a container terminated is the termination 
message (see [5]).

Per default this message needs to be explictly written to `/dev/termination-log`,
but there also is the option to use the last bit of log output as fallback source. 

4) Remove fetching of container logs from initializer

Following the focus on the exit code of the command using in the initializer job (Pod),
this commit removes the fetching of the container log.

There was only some basic JSON unmarshalling applied with no interpretation of what it contained.
This is either covered by `k6 inspect` exiting with rc=0 or should be added to the initializer
job.

If further details about the failure reason of the initalizer container was to be required,
the source should be the termination message. It could be used to enrich the `TestRun` CR to provide
a higher level of detail about the failure to the user.

[1] https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
[2] https://github.com/grafana/k6/blob/c212bd277aeedbf8395b917df11bd9be4da490a4/cmd/inspect.go#L34
[3] https://github.com/grafana/k6/blob/c212bd277aeedbf8395b917df11bd9be4da490a4/cmd/inspect.go#L64
[4] https://github.com/grafana/k6/blob/master/cmd/root.go#L90-L118
[5] https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/

Fixes: #435


frittentheke:
> Hi @frittentheke, thanks for the suggestions. I'm afraid I don't see how to make these work right now 😔
> 
> Let's go point-by-point:
> 
>     1. `EmptyDir`: I'm yet to hear anyone bringing up `readOnlyRootFilesystem` as a requirement (unless you meant this requirement is present in your setup?). Nevertheless, it's a valid point. But by your PR, it is unclear what part is supposed to read that volume. For example, right now, the logs are being read by the app directly, but I don't think there's a guarantee that `EmptyDir` of another pod (initializer) would continue to exist long enough nor that the app and initializer will be scheduled on the same node in the first place.

You are reading too much into the `EmptyDir` change. This is simply following good practice when providing the container application some temporary space (and a folder, which had to be mkdired before) to write some output into. The data in there is not required after the Job / Pod is done. Also such an `EmptyDir` is tied to the lifecycle of a Pod an cannot possibly be reused. See https://kubernetes.io/docs/concepts/storage/volumes/#emptydir . So in essence this is NOT about some shared storage or data transfer in the slightest. But seeing `mkdir` and writes to the container filesystem got me triggered.


>        Going by the rest of the description, esp. point 4, I assume you suggest to actually remove parsing the log completely. That cannot be done ATM, at all. Why we're reading the logs is mentioned in [this comment](https://github.com/grafana/k6-operator/issues/193#issuecomment-2252956906).
> 
>     2. > Remove all output redirection and grepping for error indicating strings in favor of
>        > using the exit codes from the commands to indicate success or failure reestablishing the
>        > clear contract of the inspect command to judge the provided config.
>        
>        
>        Sadly, k6 is not consistent enough around logging and exit codes both to risk such big adjustments here. If you're curious, you can see issues like [this](https://github.com/grafana/k6/issues/2804) or [this](https://github.com/grafana/k6-docs/issues/877). The current logic of a complex command in initializer did not appear out of thin air, so to say: it is meant to cover all cases possible.

I understand that the exit code is not sufficient atm. But since K6 is also a project by Grafana, there should be no reason not to fix this at the root. And if some grouping or mapping of exit codes is required one can always write up an entrypoint script handling the various exit codes and mapping them to either 0 or some other value.


>        Also, I suppose your [PR 453](https://github.com/grafana/k6-operator/pull/453) is kind of a continuation of the same idea: there is a more specific discussion over there.

Yes. More of a first step with as little code change as possible.


>     3. `TerminationMessagePolicy` on initializer: sounds interesting, but also a bit redundant, given our current reading of initializer logs.

Redundant only if you do both. My idea would be to leave the logs verbose and for the user and replace the log fetching with a fetch of the `TerminationMessage`. So to combine 2 and 3: If the exit code is zero the contract between initalizer and k6-operator would be that the TerminationMessage contains the required JSON. On any other issue (non-zero exit code), the pod log of the initializer contains all of the validation and linting, testing and whatnot output to help the user find and fix the issue (thus 4, no fetching the log and expecting it to only contain the required JSON.


>     4. > Remove fetching of container logs from initializer
>        
>        
>        Already covered above.

yorugac:
As mentioned in https://github.com/grafana/k6-operator/issues/435, I'm closing this PR. Thanks again @frittentheke :raised_hands:

Issue #455, opened 2024-08-28
Make `spec.paused` a boolean in an upcoming version of the CRD


LCaparelli:
### Feature Description

Unless there are plans to support other values and make it sort of an enum (which is also a breaking change), it probably makes more sense for `spec.paused` to be a bool. AFAICT from the code, we only use it as such anyway.

The current implementation leads to some awkwardness when using it, for example:

- in Helm charts, as part of a template, for which we need to add quotes around user input
- in Crossplane compositions, where we want to receive a bool from the user and then need to convert it to a string in order to match the current interface
- probably any abstraction built on top of this that offers the field as a bool to users

yorugac:
Thanks for the issue, @LCaparelli! Such a change makes sense, AFAIS 👍

PR #458, opened 2024-08-29
fix: don't run starter job if TestRun/K6 is paused

### LINKED ISSUE #457: `spec.paused` produces `--paused` flag on runner, but initializer always unpause it regardless
### Brief summary

While the operator correctly propagates the configuration as part of the flags for the `runner`: 

https://github.com/grafana/k6-operator/blob/0f3957c6647c81de13b26c3c4435de55c72d549d/pkg/resources/jobs/runner.go#L61-L68

It makes no checks prior to creating the starter:

https://github.com/grafana/k6-operator/blob/0f3957c6647c81de13b26c3c4435de55c72d549d/controllers/testrun_controller.go#L193-L197

Neither functions above make any checks regarding `spec.paused` AFAICT.

Since the starter job always unpauses the the runner:

https://github.com/grafana/k6-operator/blob/0f3957c6647c81de13b26c3c4435de55c72d549d/pkg/resources/containers/curl_start.go#L15-L30

We end up with `spec.paused` being ignored (technically, partially ignored because the `--paused` flag does get added to the runner arguments, but in practice it always unpauses due to how the starter behaves).

# Suggested implementation

We might solve this by just adding an `if` statement when the status is `"created"`, guarding the `StartJobs` function:

```go
	case "created":
                if(k6.IsPaused()){
                        // nothing to do. When the user updates spec.paused a new reconciliation will trigger and we'll check again
                        return ctrl.Result{}, nil 
                }
		return StartJobs(ctx, log, k6, r)
```

Of course, this implies that changing `spec.paused` after the test has started continues to have no effect, but that seems sensible to me.

0.0.16

_No response_

Anything with spec.paused set to `'true'`.

_No response_

1. Create a `TestRun` with `spec.paused` set to `'true'`
2. Wait for jobs to run

### Expected behaviour

Runner does not start the test before I set `spec.paused` to `'false'` or `null`.

### Actual behaviour

Starter behaves exactly as if `spec.paused` was set to `'false'` or `null`, and immediately starts the runner.

LCaparelli:
Fixes #457

### DIFF
--- a/api/v1alpha1/k6_types.go
+++ b/api/v1alpha1/k6_types.go
@@ -15,6 +15,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"strconv"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8stypes "k8s.io/apimachinery/pkg/types"
@@ -91,6 +93,15 @@ type TestRunSpec struct {
 	Token     string `json:"token,omitempty"`     // PLZ reserved field (for now)
 }
 
+func (k6 TestRunSpec) isPaused() bool {
+	if k6.Paused == "" {
+		return true
+	}
+
+	paused, _ := strconv.ParseBool(k6.Paused)
+	return paused
+}
+
 // K6Script describes where the script to execute the tests is found
 type K6Script struct {
 	VolumeClaim K6VolumeClaim `json:"volumeClaim,omitempty"`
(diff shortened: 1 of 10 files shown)


api/v1alpha1/k6_types.go
```diff
@@ -91,6 +93,15 @@ type TestRunSpec struct {
...
 }
 
+func (k6 TestRunSpec) isPaused() bool {
+	if k6.Paused == "" {
+		return true
+	}
+
+	paused, _ := strconv.ParseBool(k6.Paused)
+	return paused
+}
```
LCaparelli, line 103: There is a weird behavior here, where if what the string passed by the user in is not a boolean, it's interpreted as `false` by `ParseBool`. I'm just copying the existing behavior from where I extracted this logic from, but let me know if we should instead return `true`, our default if uninformed.

 We could avoid some of these shenanigans with #455 :-)

Issue #462, opened 2024-09-25
Helm Chart - Disable Leader Election if running Operator as a single Pod


hartman17:
### Feature Description

When deploying the operator using Helm there is no configurable option to disable the leader election process. This should be disabled when running the operator as a single pod, there is no need to check for leader election in this state.

### Suggested Solution (optional)

There are a couple of ways to do this: 

Add Helm config to disable leader election if  `.Values.manager.replicas == 1` 

or 

Add a `.Values.manager.leaderelection`  and allow users to pass in true/false to disable if it's not needed.

yorugac:
Solved in https://github.com/grafana/k6-operator/pull/463.

PR #463, opened 2024-09-25
Helm Chart - Enable Leader Election only if replica count greater than one

hartman17:
Addresses Issue #462 

This will enable Leader election only when the K6 Operator deployment count is greater than one.  This turns off a check that is not needed and keeps the operator logs clean.


yorugac (approved):
@hartman17, thank you for the issue and the PR! This looks great: IMHO, it is preferable to have straight-forward logic to enable / disable leader election like here.

Issue #466, opened 2024-10-01
Add securityContext to pod spec in Helm chart


chrissnell:
### Feature Description

Currently, the Helm chart allows for adding a container `securityContext` to the manager and kube-rbac-proxy containers.  We also need to add some things (like `fsGroup`) to the pod-level `securityContext`.   The chart does not provide any means to do that presently.

### Suggested Solution (optional)

This request asks for two things:

1. a new configuration variable, `global.podSecurirtyPolicy`, which can be set to a pod-level `securityPolicy` object
2. an `if` conditional that triggers an inclusion of this variable in the `deployment.yaml` template [in the pod spec](https://github.com/grafana/k6-operator/blob/2fdcbb92ee9db433ade665c503955a111c967be9/charts/k6-operator/templates/deployment.yaml#L28).

yorugac:
@chrissnell, I've just now noticed that you suggested a global variable: is there a specific reason for that? It is noticeable that `containerSecurityContext` is not global ATM.

chrissnell:
> [@chrissnell](https://github.com/chrissnell), I've just now noticed that you suggested a global variable: is there a specific reason for that? It is noticeable that `containerSecurityContext` is not global ATM.

Indeed, global variable would be ideal for this use case if they applied to all containers.  I think most users that have requirements for security contexts will need them on _all_ pods, not just individuals.  Globals would make it easy to set it for everything.

yorugac:
I see. Well, in case of k6-operator chart, there's only one deployment, with two containers. No subcharts, etc. So it seems to me there's little difference between configuration as `manager` variable and as `global` variable, except for UX becoming more confusing because of the clash with `manager.containerSecurityContext`.

chrissnell:
> I see. Well, in case of k6-operator chart, there's only one deployment, with two containers. No subcharts, etc. So it seems to me there's little difference between configuration as `manager` variable and as `global` variable, except for UX becoming more confusing because of the clash with `manager.containerSecurityContext`.

There's also the tests themselves.   I wish that the images used for the tests could be set (globally) in the Helm chart.  It's unfortunate that k6-operator users have to add all of the security context stuff and other things like custom image registries in every single `TestRun` manifest.   It would be ideal if this could all be set globally, once, in the Helm chart values.

yorugac:
>  I wish that the images used for the tests could be set (globally) in the Helm chart.

[...]

Issue #330, opened 2023-11-04
Add `CONTRIBUTING.md`


yorugac:
### Feature Description

Contributors in k6-operator has been finding their own path, and with a very good success rate, IMHO. But recently it has been raised a few times that a contributor guide would be helpful.
It's also quite noticeable that the Helm support is adding up to complexity.

### Suggested Solution (optional)

Some initial thoughts:
- it is preferable to have an issue and some discussion prior to opening a PR
- move out development section from `README.md` to `CONTRIBUTING.md` and expand it as necessary
- describe how PRs are tested
- maybe make a separation of different types of PR. Case in study: https://github.com/grafana/k6-operator/issues/328

yorugac:
The key points are now described in the https://github.com/grafana/k6-operator/blob/main/CONTRIBUTING.md

This description should cover most cases of making contributions via PRs. There are some tricky cases when PRs are more complicated than what is described, but those are rare and usually done by the maintainer. IOW, we can handle those on a case by case basis, instead of overwhelming the first version of the contributing guide with information.

Also, I think Helm PRs could have better support in general, like a test suite, perhaps. But we simply don't have that now nor do we have the bandwidth to implement it.

PR #467, opened 2024-10-02
feat: Add Helm configuration for pod security context

gonmmarques:
Hello there,

Saw this issue https://github.com/grafana/k6-operator/issues/466, and actually was also looking into having that possibility. decided to give it a try.

Tried to follow a contributing guide, but only found https://github.com/grafana/k6-operator/issues/330.

The way it is now added, by default the field is optional. Let me know also if it makes sense the way it was added (name of the key and location) and the schema definition.

I also ran helm-docs to update the README of the Helm Chart. 

I also believe `make helm-schema` needs to be run. I ran locally and noticed some other fields that I believe need to have their values changed on the `values.schema` namely containerSecurityContext on the manager and authProxy should have the additionalProperties to `true`right? Similarly the manager.image.tag.default was changed to the v0.0.016. But I was unsure how you normally go about on updating these fields.

Example 
```
global:
  podSecurityContext: {
    "runAsNonRoot": true,
    "runAsUser": 1000,
    "runAsGroup": 1000,
    "fsGroup": 1000,
    "allowPrivilegeEscalation": false,
    "seccompProfile": {
      "type": "RuntimeDefault"
    }
  }  
```  

Checked it by running locally the helm template and it rendered 
```
  spec:
      securityContext:
        allowPrivilegeEscalation: false
        fsGroup: 1000
        runAsGroup: 1000
        runAsNonRoot: true
        runAsUser: 1000
        seccompProfile:
          type: RuntimeDefault
```

Anyways, let me know if I missed something or is something wrong.
Closes: https://github.com/grafana/k6-operator/issues/466


yorugac (changes requested):
Hi @gonmmarques, apologies for the delay! I've finally gotten around to going through the PRs :sweat_smile: 

We've recently had a [bug with values schema](https://github.com/grafana/k6-operator/issues/471) in the chart: it was resolved just last week. So this PR needs rebasing to the latest main.

Also, one change request: as you can see in the [issue's](https://github.com/grafana/k6-operator/issues/466) discussion, there is a chance of misunderstanding how security contexts for test runs are configured (only via `TestRun` definition ATM). So I think its best to move the `podSecurityContext` in Helm values to `manager` object instead of `global`: to make it clearer that it impacts only the k6-operator app. WDYT?

gonmmarques:
> Hi @gonmmarques, apologies for the delay! I've finally gotten around to going through the PRs 😅
> 
> We've recently had a [bug with values schema](https://github.com/grafana/k6-operator/issues/471) in the chart: it was resolved just last week. So this PR needs rebasing to the latest main.
> 
> Also, one change request: as you can see in the [issue's](https://github.com/grafana/k6-operator/issues/466) discussion, there is a chance of misunderstanding how security contexts for test runs are configured (only via `TestRun` definition ATM). So I think its best to move the `podSecurityContext` in Helm values to `manager` object instead of `global`: to make it clearer that it impacts only the k6-operator app. WDYT?

Hello @yorugac, no worries.
Sure that makes sense. I have updated the PR, let me know if there is anything missing.

yorugac (approved):
LGTM :raised_hands: Thanks for the update, @gonmmarques!

Issue #471, opened 2024-10-03
When specifying values in the Helm chart, the installation fails.


YusukeSugiyamaTG:
### Brief summary

When using Argo CD and Helm chart for installation, specifying tolerations and affinity as follows will cause an error.

```yaml
repoURL: https://grafana.github.io/helm-charts
chart: k6-operator
targetRevision: 3.9.0
helm:
  releaseName: k6-operator
  valuesObject:
    tolerations:
      - key: "dedicated"
        operator: "Equal"
        value: "platform"
        effect: "NoSchedule"
    affinity:
      nodeAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          nodeSelectorTerms:
            - matchExpressions:
                - key: service-name
                  operator: In
                  values:
                    - platform
```

`Error details`

> Error: values don't meet the specifications of the schema(s) in the following chart(s): k6-operator: - tolerations: Invalid type. Expected: object, given: array - affinity: Additional property nodeAffinity is not allowed - affinity: Additional property podAntiAffinity is not allowed

I think that `values.schema.json` was added with this upgrade.

https://github.com/grafana/k6-operator/commit/735f8759b2030c3441e61ab8477bb1d03e5555c4#diff-5b1691e101964ac9ed5929dabe795fa6d9a722e062c88d3245e23a06b76a6317R5

```json
    "affinity": {
      "additionalProperties": false,
      "description": "affinity -- Affinity to be applied on all containers",
      "title": "affinity",
      "type": "object"
    },

    "tolerations": {
      "additionalProperties": false,
      "description": "tolerations -- Tolerations to be applied on all containers",
      "title": "tolerations",
      "type": "object"
    }
```

Here, the `additionalProperties` of `affinity` is set to `false`, so it seems that additional elements cannot be specified in `affinity`.
Regarding `tolerations`, it seems to be a pure type error.

### k6-operator version or image

0.0.17

### Helm chart version (if applicable)

3.9.0

### TestRun / PrivateLoadZone YAML

This does not apply in this case.

### Steps to reproduce the problem

The following definition and execution in the `application.yaml` file of Argo CD will cause an error.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  finalizers:
    - resources-finalizer.argocd.argoproj.io
  name: k6-operator
  namespace: argocd
spec:
  project: default
  source:
    repoURL: https://grafana.github.io/helm-charts
    chart: k6-operator
[...]

PR #479, opened 2024-10-28
Update values schema to allow resources in the proxy

pipo02mix:
Fix the error in the resource's properties for authProxy. With current configuration can not be overwritten because of
```
- authProxy.resources: limits is required
- authProxy.resources: requests is required

```

### DIFF
--- a/charts/k6-operator/values.schema.json
+++ b/charts/k6-operator/values.schema.json
@@ -73,6 +73,54 @@
         },
         "resources": {
           "additionalProperties": false,
+          "properties": {
+            "limits": {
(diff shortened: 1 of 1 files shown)


yorugac:
Hi @pipo02mix, 
Thanks for the PR! But I believe this bug was solved with the [issue](https://github.com/grafana/k6-operator/issues/471) and release of the chart [3.9.1](https://github.com/grafana/helm-charts/releases/tag/helm-k6-operator-3.9.1) just last week. Could you please switch to a newer version and check?

PR #474, opened 2024-10-04
Fixing values.yaml docs for schema generation, corrected values.schema.json

julian-perge:
With 3.9.0, the `values.schema.json` is created incorrectly from the `values.yaml`, therefore not allowing any extra env values from `manager.env` to be placed in there since it's requiring an object, but `env` is supposed to be an array of objects.

Should fix https://github.com/grafana/k6-operator/issues/471

PR #481, opened 2024-10-29
Tolerations is an array and has been updated as such in the values.schema.json

ollie-kane-CB:
Simple pull request to fix a typo.  Without this, you must use --skip-schema-validation to install the chart that uses tolerations as they're not object values; they're an array of objects.

The sample values in this project wouldn't pass validation.

### DIFF
--- a/charts/k6-operator/values.yaml
+++ b/charts/k6-operator/values.yaml
@@ -66,18 +66,18 @@ nodeSelector: {}
 # @schema
 # additionalProperties: true
 # required: false
-# type: object
+# type: array
 # @schema
-# affinity -- Affinity to be applied on all containers
-affinity: {}
+# affinity -- afinity rules to be applied on all containers
+affinity: []
 
 # @schema
 # additionalProperties: true
 # required: false
-# type: object
+# type: array
 # @schema
 # tolerations -- Tolerations to be applied on all containers
-tolerations: {}
+tolerations: []
 
 # @schema
 # required: false


yorugac:
Thanks for the PR, @ollie-kane-CB! I believe this is the same change request I've asked in https://github.com/grafana/k6-operator/pull/474 :slightly_smiling_face:

charts/k6-operator/values.yaml
```diff
@@ -61,7 +61,7 @@ affinity: {}
 # type: object
```
yorugac, line 61:
```suggestion
# type: array
```

yorugac (changes requested):
@ollie-kane-CB, there is a one-line fix that should be added; please see my comment above. Also, could you please rebase your PR against the new main branch?

yorugac:
@ollie-kane-CB, it appears you've closed the PR: is it intentional?

ollie-kane-CB:
I was fumbling with github's idiosyncracies (i'm a gitlabber, so it's like a parallel universe) and was struggling to rebase.  I closed because I don't know when I'll get around to fixing and didn't want to create noise.  I'll re-open in a few days to resolve.

yorugac:
@ollie-kane-CB, thanks for the clarification! FYI, I've fixed tolerations separately here: https://github.com/grafana/k6-operator/pull/484 in order to make a Helm release sooner. This schema bug has already remained un-fixed for way too long, IMO, so I'd like to rectify that ASAP.

Just in case, you're very welcome to pick up some other issue if you'd like to! :slightly_smiling_face:

Issue #211, opened 2023-03-30
Workload argument splitting does not respect quotes


atombender:
### Brief summary

When the operator takes the workload arguments, it splits by spaces and does not respect quotes.

### k6-operator version or image

0.0.8

### K6 YAML

```yaml
apiVersion: k6.io/v1alpha1
kind: K6
metadata:
  creationTimestamp: "2023-03-30T09:48:06Z"
  generation: 1
  labels:
    k6-workload-id: benchmark-pre-test-328cxme35q2gw
    test-id: pre-test
  name: benchmark-pre-test-328cxme35q2gw
  namespace: k6-tests
  resourceVersion: "104356441"
  uid: f46e0896-b34c-48df-9297-1d8ba9fc8885
spec:
  arguments: --tag=test_id=pre-test -o experimental-prometheus-rw -e 'QUERY=*[_type == "collection"][0..10]'
  parallelism: 10
  runner:
    env:
    - name: K6_PROMETHEUS_RW_SERVER_URL
      value: http://kps-prometheus.monitoring:9090/api/v1/write
    - name: K6_PROMETHEUS_RW_TREND_STATS
      value: p(99),p(90),p(95),avg,sum,med,max,min
    - name: K6_PROMETHEUS_RW_STALE_MARKERS
      value: "true"
    - name: INSTANCE_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.labels['job-name']
    - name: K6_JOB_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.labels['k6_cr']
    image: eu.gcr.io/sanity-cloud/xk6-sanity
    metadata: {}
    resources:
      limits:
        cpu: "3"
        memory: 8Gi
      requests:
        cpu: "3"
        memory: 8Gi
    securityContext: {}
    serviceAccountName: k6-test-runner
  script:
    configMap:
      file: test.js
      name: benchmark-pre-test-328cxme35q2gw-script
    volumeClaim:
      name: ""
  scuttle: {}
  starter:
    metadata: {}
    resources: {}
    securityContext: {}
status:
  stage: created
```

### Steps to reproduce the problem

From my spec:

```yaml
spec:
  arguments: --tag=test_id=pre-test -e 'QUERY=*[_type == "collection"][0..10]'
```

### Expected behaviour

This is the command argument list I expected:

```
    - -e
    - "QUERY=*[_type == "collection"][0..10]"
```

### Actual behaviour

[...]

Issue #162, opened 2022-11-02
Pass environment variables to k6 arguments


yorugac:
### Feature Description

At the moment, to pass environment variable as part of `arguments` string in k6-operator, one has to use external tooling, scripts, etc.
Kubernetes supports passing environment variables as [argument for pods](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments) with specific format. Its equivalent in k6 CRD would be something like this:
```
arguments: --tag pod="$(HOSTNAME)"
runner:
  env:
      - name: HOSTNAME
        value: ...
```

Attempt to use the above snippet results in k6 error due to escaped double quotes in `--tag` option:
```
time="2022-10-31T14:16:36Z" level=error msg="invalid argument \"pod=\\\"bar\\\"\" for \"--tag\" flag: parse error on line 1, column 5: bare \" in non-quoted-field"
```

Find solution to pass environment variables in `arguments` of k6 CRD.

### Suggested Solution (optional)

Adding support for passing environment variables in the above format would mean parsing `arguments` string in full and removing unnecessary escapes for `--tag`.
Alternatively, establish another approach to pass environment variables to `arguments` in k6-operator.

### Already existing or connected issues / PRs (optional)

This came up in forum question: https://community.k6.io/t/k6-operator-pass-local-environment-variable-as-test-wide-tag/5015

yorugac:
In PR https://github.com/grafana/k6-operator/pull/839, I've been figuring out a "safe" passing of env vars and added a "full" fix for initializer pod on the way. Now this form is safe to set:
```yaml
   spec:
     arguments: -e GREETING=$(MY_GREETING)
     runner:
       env:
         - name: MY_GREETING
           value: "hello world with spaces"
```
Note: the quotes in `env[].value` are for correct YAML formatting; but the value itself is stored unquoted. It is safe to pass it to `.spec.arguments` because it's passed to pods and then expanded by kubelet.

With the PR https://github.com/grafana/k6-operator/pull/839, this syntax for env vars is now substituted with `"${MY_GREETING}"` specifically for the initializer Shell script, which is then properly expanded as env var in `sh -c`. Previously, it could cause an error from initializer.

The same approach can be extended to the tags: substituting the syntax for initializer to allow format like `arguments: --tag pod=$(HOSTNAME)`. 

[...]

PR #482, opened 2024-10-30
Add a base_job_name tag to encompass all instances of a given job

chrissnell:
When outputting Prometheus metrics, k6-operator includes a `job_name` tag, but the value of this tag corresponds to the individual instance (== Pod) running a test.   So, for a job called `test`, the values of `job_name` look like `test-1`, `test-2`, etc.

This makes it very difficult to build a dashboard that displays _all jobs_ running a particular test.  We need a tag that provides the name of the job without any instance numbers.

This PR adds a `base_job_name` tag that contains only the job name.

In the example above, this PR adds a `base_job_name` tag with a value of `test`.


chrissnell:
People want test-wide tags:

https://community.grafana.com/t/k6-operator-pass-local-environment-variable-as-test-wide-tag/99639

People are struggling with bugs:

https://github.com/grafana/k6-operator/issues/162
https://github.com/grafana/k6-operator/issues/211

PR #483, opened 2024-10-30
Add a global job tag to all instances of a given job

chrissnell:
When outputting Prometheus metrics, k6-operator includes a `job_name` tag, but the value of this tag corresponds to the individual instance (== Pod) running a test.   So, for a job called `test`, the values of `job_name` look like `test-1`, `test-2`, etc.

This makes it very difficult to build a dashboard that displays _all jobs_ running a particular test.  We need a tag that provides the name of the job without any instance numbers.

People want test-wide tags:

https://community.grafana.com/t/k6-operator-pass-local-environment-variable-as-test-wide-tag/99639

People are struggling with bugs:

https://github.com/grafana/k6-operator/issues/162
https://github.com/grafana/k6-operator/issues/211

This PR adds a `base_job_name` tag that contains only the job name.

In the example above, this PR adds a `base_job_name` tag with a value of `test`.


yorugac:
Hi @chrissnell, thank for the PR! I believe this is the case when it's preferable to open an issue / forum question first and have a discussion there :sweat_smile:

The links you added are actually about troubles around quotes and env vars specifically, and the forum question was resolved. So this seems unrelated to the dashboard problem you described. Specifically:

> This makes it very difficult to build a dashboard that displays all jobs running a particular test. We need a tag that provides the name of the job without any instance numbers.

Currently, there are as many Jobs per TestRun as `parallelism` value + 2. These jobs are named with suffix attached to the name of a TestRun. So "all jobs" is basically the name of a test run CR. That information is being added to all Jobs already with the `.metadata.labels`:
```yaml
metadata:
  labels:
    k6_cr: <name>
```

So if you scrape your Kubernetes metrics to Prometheus, it should be possible to insert those labels as well. IOW, it makes sense to double-check your metrics scraping config.

For example, in Loki, I have been doing the following to get logs from all jobs per test run:
```
{cluster="...", namespace="k6-op", k6_cr="my-test"}
```

Let me know if this workaround works for you.

yorugac:
:wave: We've been having some internal discussions around metrics and labels and agreed to remove `job_name` in favor of `testrun_name`. It was described [here](https://github.com/grafana/k6-operator/issues/609). The reason is duplication of info in metrics, so a different one from what is described in this PR's description. But rationale aside, implementation-wise it amounts to a very similar thing :sweat_smile: 

@chrissnell, would you like to continue this PR for #609 ? I.e. rebase the branch and update the logic to the one described in the issue?

Issue #235, opened 2023-05-17
Update Kubebuilder go plugin


yorugac:
The operator currently uses the already deprecated `go.kubebuilder.io/v2` plugin which would be good to update to at least v3, perhaps even to v4.

Relevant docs for how to update:
- https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.3.0/#upgrade-your-go-project-from-gokubebuilderiov2-to-gokubebuilderiov3
- https://master.book.kubebuilder.io/migration/legacy/v2vsv3.html

A note on deprecation of v2:
https://master.book.kubebuilder.io/plugins/go-v2-plugin.html

yorugac:
Parts of this update were already done in PRs #212 and #217

yorugac:
The kubebuilder update was split into 2 PRs to simplify testing and reviewing:
- https://github.com/grafana/k6-operator/pull/512 - update of the app, Makefile and kustomize manifests
- https://github.com/grafana/k6-operator/pull/592 - update of the Helm chart

PR #504, opened 2024-12-10
feat: allow service without authproxy

afreyermuth98:
I want to be able to access the service without using the authproxy.
Moreover I added the way to put some custom labels and annotations on it

### DIFF
--- a/charts/k6-operator/values.yaml
+++ b/charts/k6-operator/values.yaml
@@ -115,6 +115,32 @@ prometheus:
 # required: false
 # type: object
 # @schema
+service:
+  # @schema
+  # required: false
+  # type: boolean
+  # @schema
+  # service.enabled -- enables the k6 service (default: false)
+  enabled: false
+  # @schema
+  # additionalProperties: true
+  # required: false
+  # type: object
+  # @schema
+  # service.labels -- service custom labels
+  labels: {}
+  # @schema
+  # additionalProperties: true
+  # required: false
+  # type: object
+  # @schema
+  # service.annotations -- service custom annotations
+  annotations: {}
+
+# @schema
+# required: false
+# type: object
+# @schema
 authProxy:
   # @schema
   # required: false
--- a/charts/k6-operator/values.yaml
+++ b/charts/k6-operator/values.yaml
@@ -120,8 +120,8 @@ service:
   # required: false
   # type: boolean
   # @schema
-  # service.enabled -- enables the k6 service (default: false)
-  enabled: false
+  # service.enabled -- enables the k6-operator service (default: false)
+  enabled: true
   # @schema
   # additionalProperties: true
   # required: false
--- a/charts/k6-operator/templates/service.yaml
+++ b/charts/k6-operator/templates/service.yaml
@@ -9,12 +9,12 @@ metadata:
     app.kubernetes.io/component: controller
     {{- include "k6-operator.labels" . | nindent 4 }}
     {{- include "k6-operator.customLabels" . | default "" | nindent 4 }}
-    {{- with .Values.metrics.service.annotations }}
+    {{- with .Values.service.annotations }}
     {{ toYaml . | nindent 4 }}
     {{- end }}
   annotations:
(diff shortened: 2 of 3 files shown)


yorugac:
@afreyermuth98, thanks for the PR! We don't have an issue for this yet, so could you please describe more precisely the use cases you're facing?

FYI, there are rather large structural changes in manifests happening as part of #235, and authentication and metrics will be impacted as well. So this PR can be considered only after that work is done. Given the holidays, that'll probably be in the next year. Meanwhile, if you have the time to consider how your proposal will fit in into the new structure and updated manifests, I'd appreciate it greatly! (ref. PR: https://github.com/grafana/k6-operator/pull/512, also https://github.com/kubernetes-sigs/kubebuilder/discussions/3907)

afreyermuth98:
Hello @yorugac ! Thanks for this answer :)
My needs are pretty simple : 
I have an otel collector that is scraping my pods / service depending on their annotations and I do not (want to) set auth-proxy on my k6-operator setup. But the service is created only if auth-proxy is enabled so it blocks me.
With my PR, I can force to have a service (with custom annotations) to get it scraped and collect the metrics without having to set the auth-proxy. 
Do not hesitate to reach me again if it's not clear / if something is needed :)

afreyermuth98:
@yorugac up ? :)

yorugac:
Hi @afreyermuth98, thanks for the ping! Apologies, this PR totally slipped from my radar :disappointed: 

Just for the record: the refactoring work in https://github.com/grafana/k6-operator/issues/235 has been postponed due to other, internal priorities. I'd have preferred to do that one instead, as it totally removes the auth proxy, IIRC. So once it is removed from manifests, we can remove it from the chart too. But yeah, we have what we have; let me review properly!

cc @EladAviczer, I think it's the same bug you pointed out it in #544 - please check.

charts/k6-operator/values.yaml
```diff
@@ -111,6 +111,32 @@ prometheus:
...
+# @schema
+# required: false
+# type: object
+# @schema
+service:
+  # @schema
+  # required: false
+  # type: boolean
+  # @schema
+  # service.enabled -- enables the k6 service (default: false)
```
yorugac, line 123:
```suggestion
  # service.enabled -- enables the k6-operator service (default: false)
```

charts/k6-operator/values.yaml
```diff
@@ -111,6 +111,32 @@ prometheus:
...
+  # service.labels -- service custom labels
+  labels: {}
+  # @schema
+  # additionalProperties: true
+  # required: false
+  # type: object
+  # @schema
+  # service.annotations -- service custom annotations
+  annotations: {}
+
```
yorugac, line 139: Btw, can you please describe why service would need these separately defined?
I think you mentioned it in description of your use case too, that you need annotations for scrapers. Which makes sense, but why can't they be the same as for deployment?

charts/k6-operator/values.yaml
```diff
@@ -111,6 +111,32 @@ prometheus:
...
+# required: false
+# type: object
+# @schema
+service:
+  # @schema
+  # required: false
+  # type: boolean
+  # @schema
+  # service.enabled -- enables the k6 service (default: false)
+  enabled: false
```
yorugac, line 124:
```suggestion
  enabled: true
```
Current default state for Service is actually true. Let's not change the default without a reason.

yorugac (changes requested):
This is pretty simple and overall LGTM! I've left a couple of small comments and questions -- please see below.

Also, just in case, schema can be generated with `make helm-schema` :slightly_smiling_face:

EladAviczer:
> cc @EladAviczer, I think it's the same bug you pointed out it in #544 - please check.

@yorugac Yesss :)

afreyermuth98:
@yorugac Review done !
I want to be separately adressed because I don't want to have the scraping annotation on the service AND on the pods, it would create a duplicate of scraping

yorugac (approved):
Thank you for the update, @afreyermuth98! LGTM :+1:

PR #342, opened 2023-11-21
feat: optional installation of crds

illrill:
Closes #341

Issue #505, opened 2024-12-11
Helm chart for CRDs


DanielCastronovo:
### Feature Description

Helm chart for CRDs instead on the K6 helm chart

### Suggested Solution (optional)

Helm chart for CRDs instead on the K6 helm chart

yorugac:
Hi @DanielCastronovo, sorry for the delay; it's been a holiday season. Hope you had a nice Christmas and New Year!

Could you please provide additional points as to why you wish for additional Helm chart? I've been grokking this slightly on my days off and I have some reservations about such a change. So additional clarification on the use cases, pros and cons from your perspective would be helpful.

yorugac:
Hi @DanielCastronovo, thanks for the write up!

I've been looking around for an alternative solution and it appears that there are hooks that could be utilized to help with upgrades as well. That is, [chart hooks](https://helm.sh/docs/topics/charts_hooks/) pre-upgrade. I didn't try it myself, but it sounds like there can be a Job executed one each upgrade that could update the CRDs. I.e. it should cover most of the points you raised. (except the changelog thing but we also track this in https://github.com/grafana/k6-operator/issues/344) WDYT?

Just to be clear, I'm definitely all for improving the experience with Helm chart; it's just adding a new chart to maintain would take additional time that I need to plan for. And if it's possible to have an alternative solution that won't complicate maintenance of k6-operator too much, that'd be great.

cc @mariuskimmina since you worked on a related PR 🙂

PR #506, opened 2024-12-12
fix: put k6 crds in the appropriate folder specified by Helm

mariuskimmina:
This PR aims to put CRDs folder in the appropriate folder specified by Helm which is charts/crds, so that we can skip installing CRDs whenever we want.
Please refer to: https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds

I recon that an options `installCRDs` has been added here https://github.com/grafana/k6-operator/pull/342 but it would imo be better to follow the recommended structure and let helms `skip-crds` option handle this.

I'd like to see one of these two options

* have both options of `skip-crds` and `installCRDs`
* only keep the standardized `skip-crds` and clean up `installCRDs`


mariuskimmina:
Ah, I see, this would loose the option to template our CRDs - so keeping both `installCRDs` and using `skip-crds` may not be an option.

Relevant discussion on the topic: https://github.com/cert-manager/cert-manager/issues/6179

mariuskimmina:
Perhaps https://github.com/grafana/k6-operator/issues/505 would actually be the better solution

Issue #352, opened 2023-12-13
Documentation about CR


DanielCastronovo:
### Feature Description

Provide a exhaustive list of options related to create a CR definition

yorugac:
Thanks for opening this @DanielCastronovo!
This was definitely discussed before internally so it is in the plans as part of improvement of k6-operator docs :+1:

yorugac:
Btw, apparently `quiet` option is not easy to find: https://github.com/grafana/k6-operator/issues/331#issuecomment-1880807398

yorugac:
Some options for docs generation to consider:
- https://github.com/srfrnk/crd-api-doc-gen
- https://github.com/CloudTooling/crd-api-doc-generator

yorugac:
Another question came up about CRD option `paused`:
https://community.grafana.com/t/v0-0-14-testrun-property-spec-paused-not-behaving-as-expected/124255/3

This question adds a requirement about defaults: in generated documentation, we should have default values specified so the automation must be able to extract the default values correctly as well. E.g. in Go, default value for string is `""` but for `paused` it's actually `"true"`

yorugac:
Closed with #510 

Generated docs are available here:
https://github.com/grafana/k6-operator/blob/main/docs/crd-generated.md

PR #510, opened 2024-12-15
A sample markdown doc for CRD types

yorugac:
Issue: #352 

This is an experimentation sample; unlikely to be merged like this or even in this repo.

### DIFF
--- a/docs/crd-generated.md
+++ b/docs/crd-generated.md
@@ -1,0 +1,22024 @@
+# API Reference
+
+Packages:
+
+- [k6.io/v1alpha1](#k6iov1alpha1)
+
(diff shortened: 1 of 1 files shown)


yorugac:
An update. 
This is no longer an experimentation PR: we're going to keep the generated Markdown doc in the GitHub repository for now. 
- All the main fields of the CRDs were briefly described. 
- Generation of this file is available via `make generate-crd-docs` command and should be run on each release.

In the future, we'd like to integrate this Markdown doc into the main corpus of our docs ([k6-docs](https://github.com/grafana/k6-docs) repo). But ATM, that requires more effort than we can spare; so we're keeping it here.

PR #526, opened 2025-02-04
cloud: mark abort events with "k6" origin

yorugac:
Previously, all abort events were marked with "user" origin, to distinguish from pure "k6" origin. "k6" here is not ideal either, as it's not precise, but perhaps, somewhat less misleading.


yorugac:
This is currently blocked by GCk6 work.

yorugac:
Closing this in favor of new endpoint for events - #796

PR #531, opened 2025-03-13
Helm: Enable additionalProperties for global

### LINKED ISSUE #528: JSON schema validation error when used as subchart
### Brief summary

Since introducing the values schema (#432), the k6-operator chart can no longer be used as a Helm dependency (sub chart). This is because the schema doesn't allow us to provide additional values to "our own" parent chart using e.g. `Values.global`

This has happened in many other projects, see:
(issue) https://github.com/istio/istio/issues/35495 + (fix) https://github.com/istio/istio/pull/36896
(issue) https://github.com/loft-sh/vcluster/issues/1651 + (fix) https://github.com/loft-sh/vcluster/pull/1652
(fix) https://github.com/signalfx/splunk-otel-collector-chart/pull/330

For more context see https://github.com/helm/helm/issues/10392

v0.0.19

3.11.0

N/A

_No response_

charts/k6/Chart.yaml
```yaml
apiVersion: v2
name: k6
type: application
version: 1.0.0
dependencies:
  - name: k6-operator
    version: 3.11.0
    repository: https://grafana.github.io/helm-charts/
```
charts/k6/values.yaml
```yaml
k6-operator:
  installCRDs: true
  namespace:
    create: false
  prometheus:
    enabled: true
```
env/production/values.yaml
```
global:
  cloudProvider: azure
  cloudRegion: westeurope
  clusterName: my-cluster
  environmentName: production
```

Command
```
$ helm dependency update charts/k6
$ helm template k6 charts/k6 --values env/production/values.yaml --values charts/k6/values.yaml
```

### Expected behaviour

I should be able to provide my own global values for my parent chart

### Actual behaviour

```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
k6-operator:
- global: Additional property clusterName is not allowed
- global: Additional property cloudProvider is not allowed
- global: Additional property cloudRegion is not allowed
- global: Additional property environmentName is not allowed
```

pj199x:
Fixes #528

### DIFF
--- a/charts/k6-operator/values.schema.json
+++ b/charts/k6-operator/values.schema.json
@@ -93,7 +93,7 @@
       "type": "string"
     },
     "global": {
-      "additionalProperties": false,
+      "additionalProperties": true,
       "properties": {
         "image": {
           "additionalProperties": false,
--- a/charts/k6-operator/values.yaml
+++ b/charts/k6-operator/values.yaml
@@ -11,6 +11,7 @@ nameOverride: ""
 fullnameOverride: ""
 
 # @schema
+# additionalProperties: true
 # required: false
 # type: object
 # @schema


yorugac:
My notes for posterity's sake.

According to [this doc](https://json-schema.org/draft/2020-12/json-schema-core#name-additionalproperties):
> The behavior of this keyword [additionalProperties] depends on the presence and annotation results of "properties" and "patternProperties" within the same schema object. Validation with "additionalProperties" applies only to the child values of instance names that do not appear in the annotation results of either "properties" or "patternProperties".

So this PR allows to add any field to `global` object, without impacting the validation on existing properties. But responsibility of those fields being correct is fully on user then.

To test this based on [these steps](https://github.com/grafana/k6-operator/issues/528#issue-2910137496), copy modified k6-operator chart to `charts/k6/subcharts` and use this for the main chart:
```yaml
apiVersion: v2
name: k6
type: application
version: 1.0.0
dependencies:
  - name: k6-operator
    version: 3.11.0
    repository: file://subcharts/k6-operator
```

yorugac (approved):
Thanks for the PR, @pj199x!
LGTM and by my local test, it is fixing the problem described in the issue :raised_hands: Let's try this.

PR #549, opened 2025-04-14
A write up about NDE mode in k6 OSS

yorugac:
Through the years, there has been a lot of discussion around possibility of adding [native distributed execution in k6](https://github.com/grafana/k6/issues/140), eventually culminating in this draft [PR](https://github.com/grafana/k6/pull/2816) and design doc [here](https://github.com/grafana/k6/blob/master/docs/design/020-distributed-execution-and-test-suites.md).

There are different opinions about whether k6 should support native distributed execution in the first place; with one of the counter-arguments being that k6-operator already exists and provides a form of distributed execution for OSS users if they want it. It is a valid argument despite Kubernetes requirement. But I also believe that k6-operator itself could become better as a software _with NDE implemented_. 

So here I'm adding my write up on topic. It is written a bit in a rush and I might slightly update it further in the future, but all the key points in this document were thought through months and years ago so the thought itself is not rushed at all.

Anyone is welcome to comment and bring their perspective, if it's relevant, be they from the k6 team, having k6-operator development experience or just k6-operator user experience.


docs/on-native-distributed-execution.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Why should we care?
+
+I omit the matter of the "most upvoted k6 issue", in order to focus on the matter of design of our own (Grafana) products.
+
+Given the previous section, it should come as no surprise that we in actuality have and support our own implementations of tooling for distributed mode. These implementations exist in parallel, they are partial in terms of missing parts list and tailored for the chosen specific use cases. Specifically:
+1. implementation at Grafana Cloud k6 (closed sourced). It's not supporting the whole list of 9 points yet (or not until recently? TODO: needs clarification). But most importantly, it is targetting the specific case of current implementation of GCk6 which is tied to [AWS instances](https://grafana.com/docs/grafana-cloud/testing/k6/reference/cloud-ips/#cloud-ip-list).
```
Griatch, line 18: > the whole list of 9 points 

This is the first time those 9 points are mentioned. There are not nine points listed above this anywhere. Are they in the k6 PR? Since they are important for your argument, either repeat them briefly here, or at least reference what you are talking about.

docs/on-native-distributed-execution.md
```diff
@@ -0,0 +1,79 @@
...
+
+### What about multiple regions support?
+
+Currently k6-operator solves a narrower set of use cases in comparison to GCk6 because GCk6 supports execution in [multiple regions at once](https://grafana.com/docs/grafana-cloud/testing/k6/author-run/use-load-zones/#syntax-to-set-load-zones) while k6-operator is tied to one Kubernetes cluster (also, [multi-cluster execution issue](https://github.com/grafana/k6-operator/issues/240)).
+
+There are several solutions for multi-cluster in Kubernetes nowadays but dragging one or another multi-cluster solution into k6-operator is unlikely to be a good idea. There is also hope of some form of multi-cluster support being added to Kubernetes natively but such changes move at a glacial speed and may not cover the case of k6 execution in multiple regions.
+
+Ideally, k6-operator should not depend on any one multi-cluster setup but instead be pluggable into whatever cluster and networking setup is in place. And here is where native distributed mode in k6 might come in handy too.
```
Griatch, line 69:
```suggestion
There are several solutions for multi-cluster in Kubernetes nowadays. There is also hope of some form of multi-cluster support being added to Kubernetes natively but such changes move at a glacial speed and may not cover the case of k6 execution in multiple regions.

That said, dragging one or another multi-cluster solution into k6-operator is unlikely to be a good idea. Ideally, k6-operator should not depend on any one multi-cluster setup but instead be pluggable into whatever cluster and networking setup is in place. And here is where native distributed mode in k6 might come in handy too:
```

I found these paragraphs unclear. I suggest to group them slightly differently to convey what (I think) is your point - that even though there are k8s multi-cluster solutions, cramming them into k6-operator is not a good idea. I also added a `:` at the end, since it seems the next paragraph directly references this.

Griatch (commented):
This is a good write-up, overall clear to follow and understand, even to a k8s newb like me :) I have some suggestions to clarify language/points where I found it was confusing.

docs/on-native-distributed-execution.md
```diff
@@ -0,0 +1,79 @@
...
+
+I think Kubernetes operators should not be breaking the existing status quo of abstraction layers and trying to play "services" or low-level "controllers" of other applications. Instead, it is much better for the application itself (k6 OSS) to provide a simple UX for the required functionality. Then, the operator simply uses that UX without much ado and remains simple and light-weight itself. This way the operator becomes the "glue" between the app and Kubernetes world, and the app remains self-contained and independent from Kubernetes.
+
+Curiously, it is also alluded in the NDE design doc that it's preferable to move more distributed logic to k6 level, in order to avoid implementing syncronization in a top-down fashion. This echoes my observations above: by moving the necessary logic (tooling for distributed mode) to application (k6), it is possible to keep all the higher-level abstractions clean, easy to understand and maintain. AFAIS, any other solution will likely create a mess, even if it works for some use case.
```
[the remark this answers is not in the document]
yorugac: No, it's not my interpretation. And I just checked out of curiosity:
> allude: to mention someone or something in a brief or indirect way

That bit is mentioned somewhat briefly in the NDE design doc (because the focus of that doc is slightly different), hence, "alludes".

But let's put the quote in indeed, to avoid misunderstandings.

```suggestion
Curiously, it is also alluded in the NDE design doc that top-down implementation of synchronization is part of the problem:
> Secondly, a big architectural problem of the current distributed execution solutions is that they have a push-based and top-down control mechanism. [..] This limited control mechanism is what's actually causing the need to re-implement some of the current k6 logic externally (e.g. top-down instance synchronization, setup() and teardown() handling, etc.). 

This echoes my observations above: by moving the necessary logic (tooling for distributed mode) to application (k6), it is possible to keep all the higher-level abstractions clean, easy to understand and maintain. AFAIS, any other solution will likely create a mess, even if it works for some use case.
```

docs/on-native-distributed-execution.md
```diff
@@ -0,0 +1,79 @@
...
+
+Just for completeness sake, I should point out that generally speaking, there are two ways to merge existing implementations:
+1. by moving (re-implementing, adapting, etc.) parts of GCk6 implementation into k6-operator: the problems with it are described above.
+2. by using k6-operator as GCk6 implementation, which sadly does not make much sense as:
+    - ATM, k6-operator solves a narrower set of use cases than GCk6 (more on that in the section after next).
+    - The implementation of GCk6 is rather strongly tailored for its use case on AWS and Kubernetes implementation does not map to it directly.
+
+I also think that k6-operator implementation would still benefit from simplification regardless of where it's used, for the reasons described above and in the next section.
+
+### Why NDE would help k6-operator?
```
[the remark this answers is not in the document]
yorugac: This is me writing in a hurry... :facepalm: :smile:

docs/on-native-distributed-execution.md
```diff
@@ -0,0 +1,79 @@
...
+
+The point here is that the missing parts list is not rocket science™. OTOH, the need to write that additional tooling can be seen as cumbersome. And the cost of maintenance of this tooling long-term may be too much, depending on the use case one tries to support.
+
+I'm using the term "tooling for distributed mode" to describe these additional software layers that are needed to fill the "missing parts" gaps when running k6 in distributed mode now. They can be implemented in different ways, from scripts to services. We often use the term "native distributed execution" to basically mean the implementation native to k6: that is, when most or all the tooling for distributed mode is implemented as part of k6 OSS.
+
+## Why should we care?
+
+I omit the matter of the "most upvoted k6 issue", in order to focus on the matter of design of our own (Grafana) products.
```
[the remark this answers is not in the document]
yorugac:
```suggestion
I omit the matter of the "most upvoted [k6 issue](https://github.com/grafana/k6/issues/140)", in order to focus on the matter of design of our own (Grafana) products.
```

docs/on-native-distributed-execution.md
yorugac: Actually, the "9 points" are first mentioned in the previous section "Does k6 OSS support distributed mode now?". Also named as "missing parts" and fully listed in the NDE design doc:
https://github.com/grafana/k6/blob/master/docs/design/020-distributed-execution-and-test-suites.md#missing-parts-in-the-current-state
Not the best name for referencing, but that's what we have. I'll add the link here too, to simplify the search.

```suggestion
1. implementation at Grafana Cloud k6 (closed sourced). It's not supporting the whole [list of 9 points](https://github.com/grafana/k6/blob/master/docs/design/020-distributed-execution-and-test-suites.md#missing-parts-in-the-current-state) yet (or not until recently? TODO: needs clarification). But most importantly, it is targetting the specific case of current implementation of GCk6 which is tied to [AWS instances](https://grafana.com/docs/grafana-cloud/testing/k6/reference/cloud-ips/#cloud-ip-list).
```

docs/on-native-distributed-execution.md
yorugac: :+1: plus taking it further.

```suggestion
Support of multi-cluster in Kubernetes is a long-debated problem. There is hope of some form of multi-cluster support being added to Kubernetes natively. When it is added, we definitely will be able to use it in k6-operator, but such changes move at a glacial speed and may not cover the case of k6 execution in multiple regions.

There are also several solutions for multi-cluster in Kubernetes nowadays. But dragging one or another multi-cluster solution into k6-operator is unlikely to be a good idea. Ideally, k6-operator should not depend on any one multi-cluster setup but instead be pluggable into whatever cluster and networking setup is in place. And here is where native distributed mode in k6 might come in handy too:
```

PR #585, opened 2025-05-30
Add support for multiple watched namespaces

### LINKED ISSUE #267: WATCH_NAMESPACE multiple namespaces
### Feature Description

Hi,
K6 is awesome and y'all are great!

This is a feature request to create the ability to pass multiple namespaces into WATCH_NAMESPACE. We have a few developer groups and want to keep the tests in separate namespaces.

Thanks!

I'm pretty bad at 'Go' (though I'm learning), so this might sound naive.
I think WATCH_NAMESPACE could be split on comma, and then it seems like manager.Options might be able to accept an array using `Cache.namespaces` instead of `Namespace`.

Link to manager.options relevant section:
https://github.com/kubernetes-sigs/controller-runtime/blob/40203bfeac472914f36ea35bf386326814f6f82e/pkg/manager/manager.go#L258

https://github.com/grafana/k6-operator/pull/225

chris-pinola-rf:
Adds support for a new environment variable, `WATCH_NAMESPACES`, which allows users to specify a comma-separated list of multiple Kubernetes namespaces that the k6 Operator will watch for its CRDs. The existing `WATCH_NAMESPACE` environment variable only supports a single namespace.

I added a new e2e test modeled after the existing [testrun-watch-namespace](https://github.com/grafana/k6-operator/tree/main/e2e/testrun-watch-namespace) scenario to test the new environment variable.

I verified that the new (and existing) tests pass locally on my [k3d](https://k3d.io/) cluster.

I also manually inspected the virtual cluster while the new test was running and confirmed that the `TestRun` CRD from the `invisible` namespace was not scheduled, but the other two were:
```shell
(⎈|testrun-watch-namespaces:N/A)
$ kubectl get testrun -A
NAMESPACE       NAME        STAGE      AGE   TESTRUNID
invisible       k6-sample              84s   
some-ns         k6-sample   finished   84s   
some-other-ns   k6-sample   finished   84s   
```

Fixes #267

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ Additionally, you'll need to install the following tooling:
 - [Golang](https://go.dev/doc/install)
 - [kustomize](https://kubectl.docs.kubernetes.io/installation/kustomize/)
 - [operator-sdk](https://sdk.operatorframework.io/docs/installation/): optional as most common changes can be done without it
-  - v1.31.0 is confirmed working
 
 To execute unit tests, use these commands:
 


chris-pinola-rf:
Hi @yorugac - can you please permit the test workflows to run for this PR?

yorugac (changes requested):
Thank you for this PR, @chris-pinola-rf! And thanks for adding the e2e test: multiple namespaces indeed are better as e2e. (I know the suite and the extension are not "perfect" yet but I'm happy to see it worked for you nevertheless :smiley:)

A change request is small: I ask to remove the line from Readme. Also, I think the functions in `main.go` could be refactored a bit but that's not a "requirement" for the merge, so up to you. Please see my comments with the details.

PR #592, opened 2025-06-20
Update Helm chart according to the latest kubebuilder changes

yorugac:
Issue: https://github.com/grafana/k6-operator/issues/235

This is a follow-up on PR https://github.com/grafana/k6-operator/pull/512, to ensure that Helm chart is fully up-to-date with those changes and can be upgraded successfully.


yorugac:
For record's sake.
Helm test workflow fails here, due to the problem described in #593 

But it's possible to run a successful Helm test locally with the following command:
```sh
ct install --config ./charts/ct.yaml --helm-extra-set-args "--set=namespace.create=false --set=manager.image.tag=a307fb448c9cf12b52dfffe5ea595d33954d6cff"
```

PR #593, opened 2025-06-20
CI: allow setting custom chart values in Helm test workflow

yorugac:
When there is a change both in manifests and in the code at the same time, it is possible that the chart can be tested only with a custom (not released yet) k6-operator image. In that case, it makes sense to be able to pass custom value for the image to the chart, in order for Helm test workflow to run successfully.

This PR adds manual dispatch of Helm test workflow, with an option to pass `helm_extra_args` containing a required custom value.

Case where this problem came up: https://github.com/grafana/k6-operator/pull/592

### DIFF
--- a/.github/workflows/helm-test.yaml
+++ b/.github/workflows/helm-test.yaml
@@ -11,6 +11,11 @@ on:
   pull_request:
     branches-ignore:
       - 'release-**/bundle-update'
+  workflow_dispatch:
+    inputs:
+      helm_extra_args:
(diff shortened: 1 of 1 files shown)


yorugac:
Closing this PR: this approach would cause security issues, as zizmor reported. At the same time, this problem with manifests and code changes happening at the same time should be pretty rare in practice. So for now, we can continue as we are.
Instead, perhaps, this should be additionally documented.

PR #599, opened 2025-06-30
Bind metrics server on all interfaces by default, not just loopback

chris-pinola-rf:
Only binding the metrics server to the IPv4 loopback adapter makes it impossible to scrape these metrics from outside of the controller pod.

### DIFF
--- a/charts/k6-operator/templates/deployment.yaml
+++ b/charts/k6-operator/templates/deployment.yaml
@@ -63,7 +63,7 @@ spec:
             {{- if gt (int .Values.manager.replicas) 1 }}
             - --leader-elect
             {{- end }}
-            - --metrics-bind-address=127.0.0.1:8080
+            - --metrics-bind-address=:8080
       serviceAccountName: {{ include "k6-operator.serviceAccountName" . }}
       {{- if .Values.global.image.pullSecrets }}
       imagePullSecrets:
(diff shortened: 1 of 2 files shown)


yorugac (approved):
@chris-pinola-rf, thanks for the PR!

Am I correct that you need this fix to use k6-operator properly? I'm asking to know whether I should release the updated Helm chart as a bug fix.

A comment on the topic in general. I'm OK with changing the default values, as is done here; but now I wonder if we should allow configuring these CLI arguments in the Chart's values as well :thinking: Well, I suppose someone will open the issue if they need that :smile:

chris-pinola-rf:
@yorugac No problem!

 > Am I correct that you need this fix to use k6-operator properly? I'm asking to know whether I should release the updated Helm chart as a bug fix.

I'm able to work around the issue with a Kustomize patch, but a bug fix release would be nice.

> I'm OK with changing the default values, as is done here; but now I wonder if we should allow configuring these CLI arguments in the Chart's values as well

That's a good idea. I think I would prefer to modify these arguments via the chart's values. 🤔

yorugac:
This fix was released in 3.14.1 chart today:
https://github.com/grafana/helm-charts/releases/tag/helm-k6-operator-3.14.1

PR #605, opened 2025-07-16
Fixes #604, default TerminationGracePeriodSeconds to 30s or default i…

### LINKED ISSUE #604: termination period is 0 for pods (runners in particular)
### Brief summary

Reference: https://github.com/grafana/k6-operator/blob/db5a830a8be0e40bdbd914954b8c8edb08ce96e6/pkg/resources/jobs/runner.go#L192

the value 0 prevents any log scraping (promtail, filelog otel collector mode for ex)

latest

_No response_

not relevant

_No response_

Configure any scraper and run a bench, it is likely you will randomly miss the end of the runner logs

### Expected behaviour

All logs were read

### Actual behaviour

Some logs are missing

rmannibucau:
…f pod spec is not defined


rmannibucau:
side note: cleanup mode also makes it almost impossible to use with file scrapers (out of scope of this pr)

SuperQ:
I think this may also be useful in the case where something is evicting the pods in the middle of a run. For example the cluster autoscaler.

rmannibucau:
[auto] outdated PR, autoclosing

PR #613, opened 2025-07-29
Remove default image from PLZ definition

yorugac:
This is a requirement for Grafana Cloud k6 [UI tests](https://grafana.com/docs/grafana-cloud/testing/k6/author-run/script-editor/).

The value of the default is still true, it's `grafana/k6:latest`, but the default image for PLZ tests must be set by GCk6 side; otherwise, UI test cannot be started.


yorugac:
cc @heitortsergent FYI. The format in documentation changed too.

PR #633, opened 2025-08-21
Add `portName` option to Helm chart

federicotdn:
Adds a `portName` option to the Helm chart so that the Deployment and Service port name can be configured externally.


yorugac (changes requested):
Thanks, @federicotdn! It seems to work as expected :raised_hands: 

I have one suggestion though. Could we move this to the `service.portName`? There is a `service` section already + in case we'll need to parametrize the port itself in the future, this structure would simplify addition & readability.

PR #637, opened 2025-08-29
Added feature to interpolate env vars in command args

sahilsk:
### What

This PR adds support for **environment variable interpolation** in the `arguments` field of the `TestRun` CRD.

***

### Why

When running a k6 test with multiple pods, all pods try to generate the same output files (e.g., `report.json`). This leads to file overwrites and a loss of individual pod metrics. This change allows users to use environment variables like `${POD_NAME}` to create unique filenames for each pod's output, preventing data loss. For example, `-o json=myoutput-${POD_NAME}.json` would now correctly generate `myoutput-k6-test-run-1.json`, `myoutput-k6-test-run-2.json`, etc.

Refers: https://github.com/grafana/k6-operator/issues/162
***

### How

The `jobs/runner.go` file was updated to:

1.  Collect all environment variables from the `TestRun` CRD.
2.  Use a custom function to iterate over the `arguments` string and replace any `$VAR` or `${VAR}` placeholders with their corresponding values.
3.  Append the fully interpolated arguments to the final command, ensuring the k6 binary receives the correct, unique file paths.


sahilsk:
@ppcano  @heitortsergent @yorugac   can i have your early feedback on this Draft PR please

yorugac:
Hi @sahilsk, thank you for looking into this problem! Before the code review, I have several questions here.

Firstly, have you evaluated other options for this problem? For example, other libraries or even adding a hand-written solution. In general, this flow of logic is likely the right one (get env vars for runner and pass arguments' string through some kind of substitution), but I'm not certain about relying on this library.

Also, is it possible to have a Kubernetes-style interpolation instead? I.e. with `"$(VAR)"`. Speaking of, I'd look into what Kubernetes is using for that interpolation under-the-hood: perhaps, it can be re-used.

Lastly, of course, I wonder about edge cases here: are there any?

yorugac:
In v1.6.0 release, there has been an addition of `.spec.args` field to unlock possibility to define an array of arguments to k6 and bypass the whole mess with interpretation or interpolation of variables at k6-operator level. (`.spec.arguments` continue to work; there's just another, more flexible way to set the arguments now.) 
With that, closing this PR. Thanks again for all the effort, @sahilsk!

Full details here: https://github.com/grafana/k6-operator/issues/162#issuecomment-5339044657

PR #641, opened 2025-09-07
feat: Add configurable path option for ConfigMap and VolumeClaim script sources

moko-poi:
## Summary

This PR adds support for configurable script paths in `ConfigMap` and `VolumeClaim` script sources, addressing the limitation where these options were hardcoded to use the `/test/` directory.

## Changes

- **Added `path` field** to `K6Configmap` and `K6VolumeClaim` structs
- **Updated CRD schema** to include the new optional `path` parameter
- **Enhanced documentation** to reflect the new configuration option
- **Implemented path normalization** to ensure paths always end with `/`
- **Maintained backward compatibility** by defaulting to `/test/` when path is not specified

## Motivation

Previously, `ConfigMap` and `VolumeClaim` script sources were hardcoded to mount scripts in the `/test/` directory, which:
- Didn't match the flexible behavior of `LocalFile` option
- Caused issues when users needed custom mount paths
- Limited organizational flexibility for test scripts

## Implementation Details

- The `path` field is optional and defaults to `/test/` for backward compatibility
- Path normalization ensures consistent behavior by appending `/` if missing
- Updated both parsing logic and volume mount generation
- CRD schema and generated documentation updated accordingly

## Testing

✅ **Local testing completed** with the following scenarios:

### Test Case 1: Custom Path
```yaml
apiVersion: k6.io/v1alpha1
kind: TestRun
metadata:
  name: custom-path-test
spec:
  script:
    configMap:
      name: test-scripts
      file: load-test.js
      path: /custom/scripts  # Custom path specified
```

### Test Case 2: Default Path (Backward Compatibility)
```yaml
apiVersion: k6.io/v1alpha1
kind: TestRun
metadata:
  name: default-path-test
spec:
  script:
    configMap:
      name: test-scripts
      file: load-test.js
      # No path specified - should default to /test/
```

Both test cases completed successfully, confirming:
- Custom paths work as expected
- Backward compatibility is maintained
- No breaking changes for existing configurations

## Breaking Changes

None. This is a backward-compatible enhancement.

## Related Issues

Fixes: [Issue reference if available]

## Checklist

- [x] Code follows project conventions
- [x] CRD schema updated
- [x] Documentation updated
- [x] Backward compatibility maintained
- [x] Local testing completed
- [ ] Unit tests added/updated (if applicable)

### DIFF
--- a/api/v1alpha1/testrun_types.go
+++ b/api/v1alpha1/testrun_types.go
@@ -17,6 +17,7 @@ package v1alpha1
 import (
 	"errors"
 	"path/filepath"
+	"strings"
 
 	"github.com/grafana/k6-operator/pkg/types"
 	corev1 "k8s.io/api/core/v1"
@@ -144,6 +145,9 @@ type K6VolumeClaim struct {
 	Name string `json:"name"`
 	// Name of the file to execute (.js or .tar), stored on the Volume.
 	File string `json:"file,omitempty"`
+	// Path specifies the directory path where the script file is located.
+	// If not specified, defaults to "/test/".
+	Path string `json:"path,omitempty"`
 	// ReadOnly shows whether the volume should be mounted as `readOnly`.
 	ReadOnly bool `json:"readOnly,omitempty"`
 }
@@ -154,6 +158,9 @@ type K6Configmap struct {
 	Name string `json:"name"`
 	// Name of the file to execute (.js or .tar), stored as a key in the ConfigMap.
 	File string `json:"file,omitempty"`
+	// Path specifies the directory path where the script file is located.
+	// If not specified, defaults to "/test/".
+	Path string `json:"path,omitempty"`
 }
 
 //TODO: cleanup pre-execution?
@@ -218,6 +225,13 @@ func (k6 TestRunSpec) ParseScript() (*types.Script, error) {
 			s.ReadOnly = spec.VolumeClaim.ReadOnly
 		}
 
+		if spec.VolumeClaim.Path != "" {
+			s.Path = spec.VolumeClaim.Path
+			if !strings.HasSuffix(s.Path, "/") {
+				s.Path += "/"
+			}
+		}
+
 		s.Type = "VolumeClaim"
 		return s, nil
 	}
@@ -227,6 +241,13 @@ func (k6 TestRunSpec) ParseScript() (*types.Script, error) {
 
 		if spec.ConfigMap.File != "" {
 			s.Filename = spec.ConfigMap.File
+		}
+
+		if spec.ConfigMap.Path != "" {
+			s.Path = spec.ConfigMap.Path
+			if !strings.HasSuffix(s.Path, "/") {
+				s.Path += "/"
+			}
 		}
 
 		s.Type = "ConfigMap"
--- a/config/crd/bases/k6.io_testruns.yaml
+++ b/config/crd/bases/k6.io_testruns.yaml
@@ -3916,6 +3916,8 @@ spec:
                         type: string
                       name:
                         type: string
+                      path:
+                        type: string
                     required:
                     - name
                     type: object
@@ -3926,6 +3928,8 @@ spec:
                       file:
                         type: string
                       name:
+                        type: string
+                      path:
                         type: string
                       readOnly:
                         type: boolean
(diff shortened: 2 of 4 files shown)


yorugac:
Thank you for this PR, @moko-poi! 

`LocalFile` is currently using the `Split` function:
```go
s.Path, s.Filename = filepath.Split(spec.LocalFile)
```
If there's no path, `s.Path` should be empty, it seems... Is it usable for ConfigMap and VolumeClaim, on the `file` field too? I.e. if `s.Path` is empty, default to `/test/`. Would that approach break anything? (I haven't dug into edge cases of that kind of implementation myself, so I wonder if you have :slightly_smiling_face:)

moko-poi:
Thank you for the suggestion, @yorugac! 

You're absolutely right - using filepath.Split() consistently across all script sources makes much more sense than having separate path fields.

I've updated the implementation in [5ba012c](https://github.com/grafana/k6-operator/pull/641/commits/5ba012c170b261fdaf142f08122903cd3def3b50) to follow the same approach as LocalFile:

- ConfigMap/VolumeClaim: Now use filepath.Split(spec.ConfigMap.File) and filepath.Split(spec.VolumeClaim.File)

- Backward compatibility: When s.Path is empty after splitting, it defaults to /test/ for ConfigMap/VolumeClaim

- LocalFile: Remains unchanged (no default path applied when empty)

Example usage:

```yaml

configMap:

  file: "scripts/load-test.js"  # → Path: "scripts/", Filename: "load-test.js"

  file: "load-test.js"          # → Path: "/test/", Filename: "load-test.js"

```

What do you think of this approach?

yorugac:
Thanks for the update @moko-poi. I have been testing this: AFAIS, it shouldn't break anything, so we should be good with this change.

But also, this is not super trivial, IMO, so I only hope people won't get more confused with this change :sweat_smile: 

Firstly, I don't think it's possible to have a nested structure in a ConfigMap, right? So the `file: "scripts/load-test.js"` in ConfigMap doesn't make sense and doesn't work anyway, because there is no way to create such a key. Did you find a way to make such config work, @moko-poi?

-----
For volumes, the following example has worked:

```sh
/ # ls /my-volume/ -R
/my-volume/:
bar        script.js

/my-volume/bar:
users.json
```

```js
import { SharedArray } from 'k6/data';
import { sleep } from 'k6';

const data = new SharedArray('users', function () {
  const f = JSON.parse(open('/foo/bar/users.json'));
  return f;
});

export default () => {
  const randomUser = data[Math.floor(Math.random() * data.length)];
  console.log(`${randomUser.username}, ${randomUser.password}`);
  sleep(3);
};
```

```yaml
apiVersion: k6.io/v1alpha1
kind: TestRun
metadata:
  name: k6-test-with-pvc
spec:
  script:
    volumeClaim:
      name: dynamic-pvc
      file: /foo/script.js
      readOnly: true
  parallelism: 1
```

The source of confusion can be if one assumes that they can have a `file: /foo/some-folder/script.js` and expect `/foo` as mount path and `script.js` to be in the folder `some-folder` on the volume. So the main script must always be in the root of the mount path.

moko-poi:
[c26e436](https://github.com/grafana/k6-operator/pull/641/commits/c26e436c975a9fd9318453023cb4ee07fed3c406)
Thanks for the feedback! @yorugac 
I've made the changes to address the ConfigMap path handling issue:

**Changes made**

ConfigMap behavior simplified
- Removed the misleading comment about path components in ConfigMap
- ConfigMap now uses the `file` field directly as the filename (no `filepath.Split()`)
- Mount path is always `/test` for ConfigMap (hardcoded)
- Removed the confusing documentation that suggested ConfigMap supports paths

VolumeClaim behavior maintained
- Still supports path components as intended
- Uses `filepath.Split()` to separate mount path and filename
- Mount path becomes the directory part of the `file` field

This should eliminate the confusion about ConfigMap path support while keeping the VolumeClaim functionality working as tested. The ConfigMap behavior is now much clearer - it simply takes the `file` field as a direct key name in the ConfigMap, and always mounts at `/test`.

The documentation now correctly reflects that only VolumeClaim supports path components, not ConfigMap.

yorugac (approved):
Thank you for the updates, @moko-poi!

I've re-run the tests: the previous workflows are all working and so does a new one with the path in `volumeClaim`. LGTM :raised_hands:

Issue #408, opened 2024-06-03
Error when run  test on 500 pods


volnyansky:
### Brief summary

I'm trying to run the test on 500 pods and get the error : 
 exec /usr/bin/k6: argument list too long  
I find a workaround by batching tests in 300 pods packages with the same test id

### k6-operator version or image

0.0.14

### Helm chart version (if applicable)

k6-operator-3.6.0    

### TestRun / PrivateLoadZone YAML

apiVersion: k6.io/v1alpha1
kind: TestRun
metadata:
    name: ${USERNAME}-${SCRIPT}-${BATCH}
    namespace: k6
spec:
    #number of pods to run in parallel
    parallelism: ${BATCH_PODS}
    script:
        configMap:
            name: ${USERNAME}-test-script-${BATCH}
            file: test.tar
    arguments: -o experimental-prometheus-rw --tag testid=${TESTID}
    runner:
        image: 569129334545.dkr.ecr.us-east-1.amazonaws.com/k6-robot-dev:latest
        env:
            -   name: K6_PROMETHEUS_RW_SERVER_URL
                value: "http://victoria-metrics-single-server.monitoring.svc.cluster.local:8428/api/v1/write"
            -   name: K6_PROMETHEUS_RW_TREND_STATS
                value: "count,sum,min,max,avg,med,p(90),p(95),p(99)"
            -   name: K6_BROWSER_ARGS
                value: "window-size=1920x1080,no-sandbox,disable-setuid-sandbox,allow-file-access,use-fake-device-for-media-stream,use-fake-ui-for-media-stream,use-file-for-fake-video-capture=/usr/local/assets/video.mjpeg,use-file-for-fake-audio-capture=/usr/local/assets/audio.wav"
            -   name: K6_BROWSER_TIMEOUT
                value: "45s"
            -   name: VU_ID_START
                value: "${VU_ID_START}"
        nodeSelector:
            engageli.com/role: k6-load-test
        resources:
            limits:
                cpu: "${CPU}"
                memory: ${MEMORY}Mi
            requests:
                cpu: 100m
                memory: ${MEMORY}Mi

### Steps to reproduce the problem

Run test on 500pods , number of VUs doesn't matter

### Expected behaviour

Tets runs in given number of pods

### Actual behaviour

Test crashes

volnyansky:
It think the issue is in curl max argument length , I see the following in the started pod: 
[...]

PR #642, opened 2025-09-09
Fix: Handle large hostname lists (>500 IPs)

omriKaltura:
## Problem

When running k6 tests with a large number of hostnames (more than 500 IPs), the starter job fails with a 'Linux argument list too long' error. This is a limitation of the Linux command line length.

## Solution

This PR implements a solution to handle large hostname lists by:

1. Adding a constant `maxHostnamesPerBatch = 500` to define the maximum number of hostnames per batch
2. Implementing batching logic to split the hostnames list into smaller chunks
3. Creating a loop to generate a starter job for each batch
4. Adding unique naming for each starter job when multiple batches are needed
5. Enhancing error handling to track successful job creation
6. Improving logging to show batch information and provide a summary when multiple jobs are created

## Testing

The solution has been tested with more than 500 hostnames and successfully creates multiple starter jobs to handle the large list.

### DIFF
--- a/internal/controller/k6_start.go
+++ b/internal/controller/k6_start.go
@@ -99,21 +99,54 @@ func StartJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *Te
 	}
 
 	// starter
-
-	starter := jobs.NewStarterJob(k6, hostnames)
-
-	if err = ctrl.SetControllerReference(k6, starter, r.Scheme); err != nil {
-		log.Error(err, "Failed to set controller reference for the start job")
+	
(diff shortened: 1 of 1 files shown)


yorugac:
Hi @omriKaltura, I believe this problem first came up in this issue:
https://github.com/grafana/k6-operator/issues/408

But I personally think we should solve this by [removing starter pod](https://github.com/grafana/k6-operator/issues/87) instead of creating multiple starter pods. Btw, multiple starter pods also add complexity: these pods are no longer synchronized. 
FYI, we're likely to prioritize that work with starter removal rather soon. Please watch for updates in that issue.

Issue #675, opened 2025-11-19
Enable kube-api-linter module in golangci-lint


moko-poi:
Enable kube-api-linter module in golangci-lint to enforce Kubernetes API conventions and best practices.

## Motivation
The kube-api-linter is an official linter from kubernetes-sigs that checks Kubernetes API types for common mistakes and enforces best practices based on the [Kubernetes API
Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md).

Currently, this project uses golangci-lint v2.4.0 but only has the default linters enabled (errcheck, govet, ineffassign, staticcheck, unused).

## Proposed Changes
1. Create `.custom-gcl.yml` to configure the kube-api-linter module:
```yaml
version: v2.5.0
name: golangci-lint-kube-api-linter
destination: ./bin
plugins:
  - module: 'sigs.k8s.io/kube-api-linter'
    version: 'v0.0.0-20251029102002-9992248f8813'
```

2. Create .golangci.yml to enable the linter:
```
linters-settings:
  custom:
    kubeapilinter:
      type: "module"
      description: "Kube API Linter lints Kube like APIs based on API conventions and best practices"

linters:
  enable:
    - kubeapilinter
```

3. Update CI workflow if necessary

## References
- https://github.com/kubernetes-sigs/kube-api-linter
- https://pkg.go.dev/sigs.k8s.io/kube-api-linter

yorugac:
Hi @moko-poi,
Thank you as always!

I'm afraid I don't have free capacity to process your PR within this release cycle (1.2.0) for sure. Too many internal tasks and holidays. Hope you're OK with waiting for PR feedback until 1.3.0, and LMK if not! Meanwhile, please see that CI for the PR fails:

<img width="870" height="243" alt="Image" src="https://github.com/user-attachments/assets/9e12541c-3c18-4a61-871b-74c96be7c6aa" />

----

Also, could you please elaborate what worries exactly you wish to address with the kube-api-linter. Like, is there anything specific that causes friction?

yorugac:
Closed by https://github.com/grafana/k6-operator/pull/676

PR #676, opened 2025-11-19
ci: integrate kube-api-linter into golangci-lint workflow

moko-poi:
Integrate kube-api-linter into the golangci-lint workflow to enforce Kubernetes API conventions and catch common mistakes in Kubernetes API types.

This PR addresses #675 by:
- Building golangci-lint with the kube-api-linter plugin using `golangci-lint custom`
- Updating CI workflow to use the custom-built linter
- Adding configuration file for custom plugin management

### 1. Custom golangci-lint Configuration
- **Added** `.custom-gcl.yml`: Configuration file specifying kube-api-linter plugin (v0.0.0-20251112164541-d94382a24f06) to be built into golangci-lint

### 2. Makefile Updates
- **Modified** `golangci-lint` target (Makefile:214-221): Changed from simple `go install` to custom build process
  - First installs standard golangci-lint v2.4.0
  - Then builds custom binary with plugins using `golangci-lint custom -o $(GOLANGCI_LINT) .custom-gcl.yml`
  - Ensures kube-api-linter is available in local development

### 3. CI/CD Integration
- **Modified** `.github/workflows/golangci-lint.yaml`: Replaced `golangci/golangci-lint-action` with Makefile-based approach
  - Added "Build custom golangci-lint" step to build the linter with plugins
  - Changed "lint" step to use `make lint` for consistency between local and CI environments

## Benefits

- **Standards Compliance**: Enforces [Kubernetes API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md)
automatically
- **Consistency**: Same linter configuration used in both local development and CI
- **Maintainability**: Centralized plugin configuration in `.custom-gcl.yml` for easy updates

## Testing

- Local testing: Run `make golangci-lint && make lint` to verify custom linter builds and runs correctly
- CI will automatically build and use the custom linter on all PRs and pushes

## References
- [kubernetes-sigs/kube-api-linter](https://github.com/kubernetes-sigs/kube-api-linter)
- [golangci-lint custom plugins documentation](https://golangci-lint.run/usage/configuration/#custom-plugins)

### DIFF
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -1,5 +1,5 @@
 version: v2.12.1
-name: golangci-lint
+name: golangci-lint-kube-api-linter
 destination: ./bin
 plugins:
 - module: 'sigs.k8s.io/kube-api-linter'
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -3,4 +3,4 @@ name: golangci-lint
 destination: ./bin
 plugins:
 - module: 'sigs.k8s.io/kube-api-linter'
-  version: 'v0.0.0-20251112164541-d94382a24f06'
+  version: 'v0.0.0-20260518104151-5ebe05f9440b'
--- a/.custom-gcl.yml
+++ b/.custom-gcl.yml
@@ -1,0 +1,6 @@
+version: v2.12.1
+name: golangci-lint
+destination: ./bin
+plugins:
+- module: 'sigs.k8s.io/kube-api-linter'
+  version: 'v0.0.0-20251112164541-d94382a24f06'
--- a/Makefile
+++ b/Makefile
@@ -253,9 +253,13 @@ $(ENVTEST): $(LOCALBIN)
 	$(call go-install-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest,$(ENVTEST_VERSION))
 
 .PHONY: golangci-lint
-golangci-lint: $(GOLANGCI_LINT) ## Download golangci-lint locally if necessary.
+golangci-lint: $(GOLANGCI_LINT) ## Build golangci-lint with kube-api-linter plugin.
 $(GOLANGCI_LINT): $(LOCALBIN)
-	$(call go-install-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/v2/cmd/golangci-lint,$(GOLANGCI_LINT_VERSION))
+	@if [ ! -f "$(GOLANGCI_LINT)" ]; then \
+		echo "Building golangci-lint with custom plugins..." ;\
+		GOBIN=$(LOCALBIN) go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION) ;\
+		$(LOCALBIN)/golangci-lint custom ;\
+	fi
 
 # go-install-tool will 'go install' any package with custom target and name of binary, if it doesn't exist
 # $1 - target path with name of binary
--- a/Makefile
+++ b/Makefile
@@ -254,12 +254,12 @@ $(ENVTEST): $(LOCALBIN)
 
 .PHONY: golangci-lint
 golangci-lint: $(GOLANGCI_LINT) ## Build golangci-lint with kube-api-linter plugin.
-$(GOLANGCI_LINT): $(LOCALBIN)
-	@if [ ! -f "$(GOLANGCI_LINT)" ]; then \
-		echo "Building golangci-lint with custom plugins..." ;\
-		GOBIN=$(LOCALBIN) go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION) ;\
-		$(LOCALBIN)/golangci-lint custom ;\
-	fi
+# Rebuild whenever .custom-gcl.yml changes (e.g. plugin or version bumps) so a
+# stale binary is never reused with an outdated plugin set.
+$(GOLANGCI_LINT): $(LOCALBIN) .custom-gcl.yml
+	@echo "Building golangci-lint with custom plugins..."
+	GOBIN=$(LOCALBIN) go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION)
+	$(LOCALBIN)/golangci-lint custom
 
 # go-install-tool will 'go install' any package with custom target and name of binary, if it doesn't exist
 # $1 - target path with name of binary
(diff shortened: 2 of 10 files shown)


yorugac:
@moko-poi, hello! Would you like to clean up / continue with this PR? As mentioned in the issue, it had [failing CI](https://github.com/grafana/k6-operator/issues/675#issuecomment-3639030426) back then and by now, it also needs a rebase :sweat_smile: 
PS just in case, a forewarning: all the commits need to be signed now! New policy; can't be helped.

moko-poi:
Hi @yorugac, thanks for the nudge and for the heads-up about the signing policy. I've gone ahead and cleaned this up.

I rebased the branch onto the latest main and resolved the conflicts, keeping golangci-lint at the current v2.12.1 used on main. All commits are now signed as well.

While doing this I found the real reason the lint job was failing: the previous version only built golangci-lint with the kube-api-linter plugin but never actually enabled it, since there was no .golangci.yml to register the module linter. Once I enabled it properly, the linter reported 118 violations across the existing v1alpha1 API types, which is what was breaking CI.

To avoid turning this into one huge change, I went with a staged rollout. I added a .golangci.yml that scopes kubeapilinter to the api/ directory only, so it doesn't flag ordinary internal Go structs under pkg/. The rules with a large number of pre-existing violations (commentstart, optionalorrequired, optionalfields, arrayofstruct, ssatags) are disabled for now and can be enabled incrementally in follow-ups. I did fix the smaller, more meaningful set right away: the statusoptional and conditions rules. That meant marking the TestRunStatus and PrivateLoadZoneStatus fields as optional and adding the standard listType/patchStrategy markers to the Conditions fields, and I regenerated the CRDs accordingly.

make lint now passes cleanly. Let me know if you'd prefer a different scope for the first step, or if you'd rather see more rules enabled in this PR before merging. Happy to adjust.

Makefile
```diff
@@ -253,9 +253,13 @@ $(ENVTEST): $(LOCALBIN)
...
 .PHONY: golangci-lint
-golangci-lint: $(GOLANGCI_LINT) ## Download golangci-lint locally if necessary.
+golangci-lint: $(GOLANGCI_LINT) ## Build golangci-lint with kube-api-linter plugin.
 $(GOLANGCI_LINT): $(LOCALBIN)
-	$(call go-install-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/v2/cmd/golangci-lint,$(GOLANGCI_LINT_VERSION))
+	@if [ ! -f "$(GOLANGCI_LINT)" ]; then \
+		echo "Building golangci-lint with custom plugins..." ;\
+		GOBIN=$(LOCALBIN) go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION) ;\
+		$(LOCALBIN)/golangci-lint custom ;\
+	fi
```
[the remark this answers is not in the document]
moko-poi: Good catch. Fixed in 596e0e2: the target now depends on .custom-gcl.yml and drops the `if [ ! -f ]` guard, so the binary is rebuilt whenever the plugin config or version changes instead of being skipped when a stale binary is present.

config/crd/bases/k6.io_testruns.yaml
```diff
@@ -6203,6 +6203,9 @@ spec:
                   - type
                   type: object
                 type: array
+                x-kubernetes-list-map-keys:
+                - type
+                x-kubernetes-list-type: map
```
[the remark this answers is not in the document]
moko-poi: Fixed in 596e0e2. I synced the same listType=map / listMapKeys=[type] markers for status.conditions into the Helm chart CRD templates (charts/k6-operator/templates/crds/testrun.yaml and plz.yaml), so Helm installs now expose the same schema as config/crd/bases.

.custom-gcl.yml
```diff
@@ -0,0 +1,6 @@
+version: v2.12.1
+name: golangci-lint
+destination: ./bin
+plugins:
+- module: 'sigs.k8s.io/kube-api-linter'
+  version: 'v0.0.0-20251112164541-d94382a24f06'
```
yorugac: This seems pretty old?
Looking at: https://pkg.go.dev/sigs.k8s.io/kube-api-linter?tab=versions

Makefile
```diff
@@ -253,9 +253,13 @@ $(ENVTEST): $(LOCALBIN)
 	$(call go-install-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest,$(ENVTEST_VERSION))
 
 .PHONY: golangci-lint
-golangci-lint: $(GOLANGCI_LINT) ## Download golangci-lint locally if necessary.
-$(GOLANGCI_LINT): $(LOCALBIN)
-	$(call go-install-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/v2/cmd/golangci-lint,$(GOLANGCI_LINT_VERSION))
+golangci-lint: $(GOLANGCI_LINT) ## Build golangci-lint with kube-api-linter plugin.
+# Rebuild whenever .custom-gcl.yml changes (e.g. plugin or version bumps) so a
+# stale binary is never reused with an outdated plugin set.
+$(GOLANGCI_LINT): $(LOCALBIN) .custom-gcl.yml
```
yorugac, line 264: Can this be written re-using `go-install-tool`?

yorugac (changes requested):
@moko-poi, thank you. I kind of hoped that kube-api-linter would have some kind of release at least, but it seems like it's quite a bit more involved to add it 😅 I agree with the decisions on cutting scope that you made :+1: 
 

I have 2 request changes: please see the comments. Also, the lint step in CI doesn't work yet:

Image

From: https://github.com/grafana/k6-operator/actions/runs/28075683570/job/84513210187 

I didn't yet try to run this locally, but is it trying to overwrite the currently executing binary? At least that's how the error looks like.

.custom-gcl.yml
moko-poi: @yorugac Good point. I bumped it to the latest available pseudo-version (2026-05-18)

Makefile
moko-poi: @yorugac Done. It turned out the lint job was failing because `golangci-lint custom` was writing over the running bootstrap binary (text file busy on Linux), so following the kube-api-linter docs (https://github.com/kubernetes-sigs/kube-api-linter?tab=readme-ov-file#golangci-lint-module) I now give the custom binary a distinct name (`golangci-lint-kube-api-linter`). That removes the collision and let me reuse `go-install-tool` for the bootstrap install as you suggested.

yorugac (approved):
@moko-poi, thank you for the updates and clarity of explanations! I think Makefile is cleaner now and it's easier to follow the logic 🙌

Issue #667, opened 2025-11-07
Helm Chart: RBAC Authorization Failures When `rbac.namespaced=true`


bcrisp4:
### Brief summary

When deploying the k6-operator chart with `rbac.namespaced=true`, the operator fails to function due to missing RBAC permissions. The chart does not create equivalent namespace-scoped RBAC resources (Role/RoleBinding) when this option is enabled, causing authorization failures for all operator functionality.

### k6-operator version or image

1.0.0

### Helm chart version (if applicable)

4.0.0

### TestRun / PrivateLoadZone YAML

N/A

### Steps to reproduce the problem

1. Install the k6-operator Helm chart with namespaced RBAC enabled
2. Check the operator pod logs
3. Observe authorization errors in the controller logs indicating the ServiceAccount lacks necessary permissions

### Expected behaviour

When `rbac.namespaced=true` is set in values.yaml, the Helm chart should:

- Create namespace-scoped Role resources with the same permissions as the cluster-scoped ClusterRole resources
- Create namespace-scoped RoleBinding resources that bind these roles to the operator's ServiceAccount
- Allow the operator to function normally within its namespace without requiring cluster-wide permissions
This would enable the operator to run in restricted environments where cluster-scoped RBAC permissions are not available or not desired.

### Actual behaviour

When `rbac.namespaced=true` is set:

- The chart only prevents the creation of cluster-scoped RBAC resources (ClusterRole/ClusterRoleBinding)
- No equivalent namespace-scoped RBAC resources are created
- The operator's ServiceAccount has no permissions to access required Kubernetes resources
- The operator fails with authorization errors such as:

```text
{"level":"error","ts":"2025-11-07T20:44:57Z","logger":"controller-runtime.cache.UnhandledError","msg":"Failed to watch","reflector":"pkg/mod/k8s.io/client-go@v0.34.0/tools/cache/reflector.go:290","type":"*v1alpha1.PrivateLoadZone","error":"failed to list *v1alpha1.PrivateLoadZone: privateloadzones.k6.io is forbidden: User \"system:serviceaccount:k6-operator:k6-operator-controller\" cannot list resource \"privateloadzones\" in API group \"k6.io\" in the namespace \"k6-operator\""}
```

yorugac:
Hi @bcrisp4, thanks for the issue.

It seems like the Role is created but it's the one only for leader election and it doesn't have all the necessary permissions otherwise:
```sh
[...]

PR #698, opened 2025-12-19
Create `Roles` on `rbac.namespaced=true`

yorugac:
Fixes: https://github.com/grafana/k6-operator/issues/667 

Changes:
- `role*.yaml` were renamed to `leaderRole*.yaml`
- `clusterRole*.yaml` were renamed to `role*.yaml`
- newly renamed `role*.yaml` received new logic to cover the case of `rbac.namespaced=true` correctly: Role and RoleBinding will be created, with the same permissions as ClusterRole and ClusterRoleBinding have (in non-namespaced mode).
    - Note: `metrics-reader` can be only `ClusterRole` as it configures `nonResrouceURLs` rule for `/metrics` endpoint. There is no equivalent for that as namespaced `Role`.  So `metircs-reader` is simply not created with `rbac.namespaced=true`.
- if `rbac.namespaced=true`, manager gets `WATCH_NAMESPACE` env var pointing to its own namespace, where all Roles are created. 
    - Note: this logic is not checking for potential duplicate definitions of `WATCH_NAMESPACE` in `manager.env` values. Checking for conflicts in this case is considered a responsibility of a user to set up correctly, at the moment. Basically, don't use `WATCH_NAMESPACE(S)` env vars in namespaced mode.


charts/k6-operator/templates/role.yaml
```diff
@@ -32,9 +70,193 @@ rules:
...
+  annotations:
+    {{- include "k6-operator.customAnnotations" . | default "" | nindent 4 }}
+rules:
+- nonResourceURLs:
+  - /metrics
   verbs:
-    - create
-    - patch
\ No newline at end of file
+  - get
```
mshuler-niyamit: Error: INSTALLATION FAILED: 1 error occurred:
        * Role.rbac.authorization.k8s.io "k6-operator-metrics-reader" is invalid: rules[0].nonResourceURLs: Invalid value: ["/metrics"]: namespaced rules cannot apply to non-resource URLs

```suggestion
{{- if not .Values.rbac.namespaced }}
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: {{ include "k6-operator.fullname" . }}-metrics-reader
  {{- if .Values.rbac.namespaced }}
  namespace: {{- include "k6-operator.namespace" . }}
  {{- end }}
  labels:
    {{- include "k6-operator.labels" . | nindent 4 }}
    {{- include "k6-operator.customLabels" . | default "" | nindent 4 }}
  annotations:
    {{- include "k6-operator.customAnnotations" . | default "" | nindent 4 }}
rules:
- nonResourceURLs:
  - /metrics
  verbs:
  - get
{{- end }}
```

mshuler-niyamit:
Additionally, it would be good to have the WATCH_NAMESPACE env conditionally injected with rbac.namespaced

```yaml
          {{- if or .Values.manager.env .Values.rbac.namespaced }}
          env:
            {{- if .Values.rbac.namespaced }}
            - name: WATCH_NAMESPACE
              value: {{ include "k6-operator.namespace" . }}
            {{- end }}
```

yorugac:
> Additionally, it would be good to have the WATCH_NAMESPACE env conditionally injected with rbac.namespaced

Well, namespaced mode won't even work without it. So it has to be injected.
This logic could be further improved, I suppose :thinking: But we'll wait until there is a feature request with a use case for that. As a fix, this should be sufficient.

charts/k6-operator/templates/role.yaml
yorugac: This was fixed.

PR #755, opened 2026-03-31
Add `cloud` section in TestRun CRD

### LINKED ISSUE #668: Add a `cloud` section to `TestRun` CRD
### Feature Description

So far, we have been inserting bits specific for Grafana Cloud k6 into `TestRun` definition rather arbitrarily. For example, we have [2 fields](https://github.com/grafana/k6-operator/blob/2ce1ed83ac04628fa0ce758abc7662663801b730/api/v1alpha1/testrun_types.go#L127-L131) at a top-level of `.spec` which are reserved for PLZ and not meant to be used by anything else. 

Additionally, k6 OSS considers `--out cloud` format [deprecated](https://github.com/grafana/k6/releases/tag/v0.54.0), so cloud output in the `TestRun` should try to support that direction and new format. Problem: currently, [cloud output in k6-operator](https://grafana.com/docs/k6/latest/set-up/set-up-distributed-k6/usage/k6-operator-to-gck6/#cloud-output) depends on passing `--out cloud` as part of `.spec.arguments`, so it's relying on a deprecated format of the command.

Finally, in the future, there can be other GCk6 specific parameters we might need to pass to the `TestRun` CRD.

To improve structure of the `TestRun` CRD and facilitate migration to the `k6 cloud run --local-execution`, implement the following changes:
```yaml
apiVersion: k6.io/v1alpha1
kind: TestRun
metadata:
  name: test
spec:
  cloud: # GCk6 reserved fields
    stream: true # a new way to indicate a cloud output mode in k6-operator
    token: .. # for PLZ (for now)
    testRunId: ... # for PLZ
```

As a second iteration, it might make sense to consider `.spec.cloud.token` as token for cloud output tests as well.

**Note** This must be a backwards-compatible change. So basically, k6-operator will need to support 2 possible ways to indicate a cloud output test, as well as support both formats of PLZ tests (with `cloud` section and without it). Revisit maintaining both spec closer to k6-operator v2.0.0.

Not yet set in stone, but this might be a prerequisite for upgrading `TestRun` CRD version #623

yorugac:
Fixes: #668 

The expected behavior:
 - `--out cloud` alone: works (runs as `k6 run --out cloud`) and logs deprecation warning
 - `cloud.stream` alone: works (runs as `k6 cloud run --local-execution`)
 - Both together: validation error suggesting to use only `cloud.stream`

The below was pulled out into a separate [PR](https://github.com/grafana/k6-operator/pull/774).
~~Additionally~~ 
- ~~Adding `-race` to `make test`~~
- ~~Adding some unit tests on related code parts~~


yorugac:
Due to org's signature limitations, closing in favor of new PR: https://github.com/grafana/k6-operator/pull/845

PR #773, opened 2026-04-06
Improve unit tests

yorugac:
Another maintenance PR. Changes:

- Refactor outdated monstrosity in `runner_test.go`
- Add some missing unit tests, esp. around `volumeClaim`
- Comment on `initializer.disabled` behaviour
    - With this field, we basically introduced ad-hoc, conditional logic, which may seem convoluted on the first glance. Comment is propagated to generated CRD docs and hopefully, should help understanding a bit.

### DIFF
--- a/pkg/types/script_test.go
+++ b/pkg/types/script_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	deep "github.com/go-test/deep"
+	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/types/script_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+import (
+	"testing"
+
+	deep "github.com/go-test/deep"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func Test_Script_Volume(t *testing.T) {
```
yorugac, line 11: Test cases for `Volume` were moved from the refactored `runner_test.go` to `script_test.go`.

Issue #742, opened 2026-03-17
PLZ: refine env vars and add new CLI arguments


yorugac:
### Feature Description

Current definition and capabilities of PLZ test runs are minimal in comparison to public GCk6 tests. We're going to polish it a bit, so that PLZ becomes more similar to other kinds of tests. Specifically:
1. The `.spec.arguments` definition should be expanded with some additional flags to k6 CLI: `--no-summary`, `--verbose` (TODO: clarify this one)
1. User-defined env vars must be passed with `-e` CLI flag
1. The call to `/loadtests/v4/test_runs(%s)?$select=id,run_status,k8s_load_zones_config,k6_runtime_config` must also include `,load_zone_distribution` and the format of expected response is now this:
    ```json
    {
      "k8s_load_zones_config": {
        "instance_count": <int>,
        "k6_archive_temp_public_url": <str>,
        "load_runner_image": <str>,
  
        "environment": <map>, # user-defined
  
        "cli_flags": { # user-defined
          "tags": <map>,
          "user_agent": <str>,
          "blacklist_ips": [<str>], 
          "blocked_hostnames": [<str>],
          "include_system_env_variables": bool,
        },
        "gck6_env_vars": { # defined by GCk6
          "K6_CLOUD_TOKEN": "token",
          "K6_LOG_OUTPUT": "value", 
        },
      },
      load_zone_distribution: {
            "label": {"loadZone": <str>, "percent": <int>}
      }
    ```
1. `cli_flags` must all be passed to the `.spec.arguments` as [k6 tags](https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/#tags), `--user-agent`, `--blacklist-ip` and `--block-hostnames` CLI args respectively.
1. `gck6_env_vars` must be set as env vars for the runner
    - NOTE: `K6_LOG_OUTPUT` format is pending; we need to ensure we don't have a degradation here.
1. As described [here](https://grafana.com/docs/grafana-cloud/testing/k6/author-run/cloud-scripting-extras/cloud-execution-context-variables/), we need to set 4 additional env vars as part of user-defined env vars, via `-e` tag.
    - three of them can be set as part of `.spec.arguments`: `K6_CLOUDRUN_LOAD_ZONE` (taken from `distribution` field), `K6_CLOUDRUN_DISTRIBUTION` (also from `distribution` field?), `K6_CLOUDRUN_TEST_RUN_ID`.
[...]

PR #777, opened 2026-04-09
[feat] Transmit new environment variables in the k6 Operator when running the k6 binary

### LINKED ISSUE #747: PLZ: k6Secrets - Transmit new environment variables in the k6 Operator when running the k6 binary
### Feature Description

https://github.com/grafana/k6-cloud/issues/4267

The k6 operator should start setting new environment variables when running k6 binary:
- two key data: the `test_run_token` and `secret_decryption_url`
- other information for the secret subsystem of k6 to be able to hit that URL and process correctly its response
- `SECRETS_SOURCE` set to `url`

_No response_

_No response_

vortegatorres:
https://github.com/grafana/k6-cloud/issues/4267

- Extend the GetTestRunData query to fetch `test_run_token` and `secrets_config` from the metrics-api endpoint.                                                                                                                                                          
- When secrets_config is present, inject `K6_SECRET_SOURCE=url`, `K6_SECRET_SOURCE_URL_URL_TEMPLATE`, `K6_SECRET_SOURCE_URL_RESPONSE_PATH`, and (if a token exists) `K6_SECRET_SOURCE_URL_HEADER_AUTHORIZATION` into the runner pod's environment, enabling k6 to decrypt secrets via the k6 Cloud secrets API.                                                                                                                                                                                                                                              
- Env-var-only approach (no `--secret-source url` CLI flag) ensures backward compatibility: runner images with k6 < v1.5.0 silently ignore the unknown vars and continue working normally.
   

Fixes: https://github.com/grafana/k6-operator/issues/747

### DIFF
--- a/internal/controller/testrun_controller.go
+++ b/internal/controller/testrun_controller.go
@@ -327,13 +327,9 @@ func (r *TestRunReconciler) reconcile(ctx context.Context, req ctrl.Request, log
 			}
 		}
 
-		// If this is a non-PLZ cloud test run, try to finalize it.
-		// PLZ test runs are finalized by k6's own cloud output when the scheduler
-		// stops; the operator's FinishTestRun call (POST /v1/tests/{id}) returns
-		// 404 for PLZ runs and must be skipped to avoid getting stuck here.
+		// If this is a cloud test run in any mode, try to finalize it.
 		if v1alpha1.IsTrue(k6, v1alpha1.CloudTestRun) &&
-			v1alpha1.IsFalse(k6, v1alpha1.CloudTestRunFinalized) &&
-			v1alpha1.IsFalse(k6, v1alpha1.CloudPLZTestRun) {
+			v1alpha1.IsFalse(k6, v1alpha1.CloudTestRunFinalized) {
 
 			// If TestRunRunning has just been updated, wait for a bit before
 			// acting, to avoid race condition between different reconcile loops.
--- a/internal/controller/testrun_controller.go
+++ b/internal/controller/testrun_controller.go
@@ -327,9 +327,13 @@ func (r *TestRunReconciler) reconcile(ctx context.Context, req ctrl.Request, log
 			}
 		}
 
-		// If this is a cloud test run in any mode, try to finalize it.
+		// If this is a non-PLZ cloud test run, try to finalize it.
+		// PLZ test runs are finalized by k6's own cloud output when the scheduler
+		// stops; the operator's FinishTestRun call (POST /v1/tests/{id}) returns
+		// 404 for PLZ runs and must be skipped to avoid getting stuck here.
 		if v1alpha1.IsTrue(k6, v1alpha1.CloudTestRun) &&
-			v1alpha1.IsFalse(k6, v1alpha1.CloudTestRunFinalized) {
+			v1alpha1.IsFalse(k6, v1alpha1.CloudTestRunFinalized) &&
+			v1alpha1.IsFalse(k6, v1alpha1.CloudPLZTestRun) {
 
 			// If TestRunRunning has just been updated, wait for a bit before
 			// acting, to avoid race condition between different reconcile loops.
--- a/pkg/cloud/types.go
+++ b/pkg/cloud/types.go
@@ -57,6 +57,12 @@ type testRunList struct {
 	} `json:"object"`
 }
 
+// SecretsConfig holds the secrets configuration returned by k6 Cloud.
+type SecretsConfig struct {
+	Endpoint     string `json:"endpoint"`
+	ResponsePath string `json:"response_path"`
+}
+
 // TestRunData holds the output from /loadtests/v4/test_runs(%s)
 type TestRunData struct {
 	TestRunId     int `json:"id"`
@@ -64,6 +70,8 @@ type TestRunData struct {
 	LZConfig      `json:"k8s_load_zones_config"`
 	RunStatus     cloudapi.RunStatus `json:"run_status"`
 	RuntimeConfig cloudapi.Config    `json:"k6_runtime_config"`
+	TestRunToken  string             `json:"test_run_token,omitempty"`
+	SecretsConfig *SecretsConfig     `json:"secrets_config,omitempty"`
 }
 
 type LZConfig struct {
--- a/pkg/cloud/types.go
+++ b/pkg/cloud/types.go
@@ -63,6 +63,13 @@ type SecretsConfig struct {
 	ResponsePath string `json:"response_path"`
 }
 
+const (
+	secretSourceEnvVar      = "K6_SECRET_SOURCE"
+	secretSourceURLTemplate = "K6_SECRET_SOURCE_URL_URL_TEMPLATE"
+	secretSourceURLRespPath = "K6_SECRET_SOURCE_URL_RESPONSE_PATH"
+	secretSourceURLAuthKey  = "K6_SECRET_SOURCE_URL_HEADER_AUTHORIZATION"
+)
+
 // TestRunData holds the output from /loadtests/v4/test_runs(%s)
 type TestRunData struct {
 	TestRunId     int `json:"id"`
@@ -100,6 +107,26 @@ func (lz *LZConfig) EnvVars() []corev1.EnvVar {
 		return ev[i].Name < ev[j].Name
 	})
 
+	return ev
+}
+
+// SecretsEnvVars returns the env vars required by the k6 URL secret source.
+// Returns nil when no secrets configuration is present.
+func (trd *TestRunData) SecretsEnvVars() []corev1.EnvVar {
+	if trd.SecretsConfig == nil {
+		return nil
+	}
+	ev := []corev1.EnvVar{
+		{Name: secretSourceEnvVar, Value: "url"},
+		{Name: secretSourceURLTemplate, Value: trd.SecretsConfig.Endpoint},
+		{Name: secretSourceURLRespPath, Value: trd.SecretsConfig.ResponsePath},
+	}
+	if trd.TestRunToken != "" {
+		ev = append(ev, corev1.EnvVar{
+			Name:  secretSourceURLAuthKey,
+			Value: "Bearer " + trd.TestRunToken,
+		})
+	}
 	return ev
 }
 
--- a/pkg/cloud/types.go
+++ b/pkg/cloud/types.go
@@ -77,7 +77,8 @@ type TestRunData struct {
 	LZConfig      `json:"k8s_load_zones_config"`
 	RunStatus     cloudapi.RunStatus `json:"run_status"`
 	RuntimeConfig cloudapi.Config    `json:"k6_runtime_config"`
-	TestRunToken  string             `json:"test_run_token,omitempty"`
+	// SecretsToken is a short-lived, test-run-scoped token for read-only access to secrets.
+	SecretsToken  string             `json:"test_run_token,omitempty"`
 	SecretsConfig *SecretsConfig     `json:"secrets_config,omitempty"`
 }
 
@@ -121,10 +122,10 @@ func (trd *TestRunData) SecretsEnvVars() []corev1.EnvVar {
 		{Name: secretSourceURLTemplate, Value: trd.SecretsConfig.Endpoint},
 		{Name: secretSourceURLRespPath, Value: trd.SecretsConfig.ResponsePath},
 	}
-	if trd.TestRunToken != "" {
+	if trd.SecretsToken != "" {
 		ev = append(ev, corev1.EnvVar{
 			Name:  secretSourceURLAuthKey,
-			Value: "Bearer " + trd.TestRunToken,
+			Value: "Bearer " + trd.SecretsToken,
 		})
 	}
 	return ev
--- a/pkg/cloud/types_test.go
+++ b/pkg/cloud/types_test.go
@@ -110,12 +110,6 @@ func TestTestRunData_SecretsEnvVars(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			got := tt.trd.SecretsEnvVars()
-			if tt.expected == nil {
-				if got != nil {
-					t.Errorf("SecretsEnvVars() = %+v, want nil", got)
-				}
-				return
-			}
 			if len(got) != len(tt.expected) {
 				t.Fatalf("SecretsEnvVars() len = %d, want %d; got %+v", len(got), len(tt.expected), got)
 			}
--- a/pkg/cloud/types_test.go
+++ b/pkg/cloud/types_test.go
@@ -80,17 +80,6 @@ func TestTestRunData_SecretsEnvVars(t *testing.T) {
 			expected: nil,
 		},
 		{
-			name: "secrets config without token omits auth header",
-			trd: TestRunData{
-				SecretsConfig: &SecretsConfig{Endpoint: someEndpoint, ResponsePath: someRespPath},
-			},
-			expected: []corev1.EnvVar{
-				{Name: secretSourceEnvVar, Value: "url"},
-				{Name: secretSourceURLTemplate, Value: someEndpoint},
-				{Name: secretSourceURLRespPath, Value: someRespPath},
-			},
-		},
-		{
 			name: "secrets config with token includes auth header",
 			trd: TestRunData{
 				SecretsConfig: &SecretsConfig{Endpoint: someEndpoint, ResponsePath: someRespPath},
--- a/pkg/cloud/types_test.go
+++ b/pkg/cloud/types_test.go
@@ -83,7 +83,7 @@ func TestTestRunData_SecretsEnvVars(t *testing.T) {
 			name: "secrets config with token includes auth header",
 			trd: TestRunData{
 				SecretsConfig: &SecretsConfig{Endpoint: someEndpoint, ResponsePath: someRespPath},
-				TestRunToken:  someToken,
+				SecretsToken:  someToken,
 			},
 			expected: []corev1.EnvVar{
 				{Name: secretSourceEnvVar, Value: "url"},
--- a/pkg/cloud/types_test.go
+++ b/pkg/cloud/types_test.go
@@ -3,6 +3,8 @@ package cloud
 import (
 	"encoding/json"
 	"testing"
+
+	corev1 "k8s.io/api/core/v1"
 )
 
 func TestInspectOutput_TestNameAndProjectID(t *testing.T) {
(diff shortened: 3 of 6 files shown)


pkg/plz/worker.go
```diff
@@ -177,6 +177,20 @@ func (w *PLZWorker) complete(tr *v1alpha1.TestRun, trData *cloud.TestRunData) {
...
+		envVars = append(envVars,
+			corev1.EnvVar{Name: "K6_SECRET_SOURCE", Value: "url"},
+			corev1.EnvVar{Name: "K6_SECRET_SOURCE_URL_URL_TEMPLATE", Value: trData.SecretsConfig.Endpoint},
+			corev1.EnvVar{Name: "K6_SECRET_SOURCE_URL_RESPONSE_PATH", Value: trData.SecretsConfig.ResponsePath},
+		)
+		if trData.TestRunToken != "" {
+			envVars = append(envVars, corev1.EnvVar{
+				Name:  "K6_SECRET_SOURCE_URL_HEADER_AUTHORIZATION",
+				Value: "Bearer " + trData.TestRunToken,
+			})
```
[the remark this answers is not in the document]
vortegatorres: Fixed in https://github.com/grafana/k6-operator/pull/777/changes/847633c412a4edd52eece5c7b70f074b841732c8

pkg/plz/worker_test.go
```diff
@@ -227,6 +231,29 @@ func Test_complete_correctDefinitionOfTestRun(t *testing.T) {
...
+		corev1.EnvVar{Name: "K6_SECRET_SOURCE_URL_HEADER_AUTHORIZATION", Value: "Bearer " + someTestRunToken},
+	)
+
+	secretsWithoutTokenTestRun = cloudFieldsTestRun
+	secretsWithoutTokenTestRun.Spec.Runner.Env = append(
+		append([]corev1.EnvVar{}, defaultTestRun.Spec.Runner.Env...),
+		corev1.EnvVar{Name: "K6_SECRET_SOURCE", Value: "url"},
+		corev1.EnvVar{Name: "K6_SECRET_SOURCE_URL_URL_TEMPLATE", Value: someSecretsConfig.Endpoint},
+		corev1.EnvVar{Name: "K6_SECRET_SOURCE_URL_RESPONSE_PATH", Value: someSecretsConfig.ResponsePath},
+	)
```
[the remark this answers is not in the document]
vortegatorres: Fixed in https://github.com/grafana/k6-operator/pull/777/changes/847633c412a4edd52eece5c7b70f074b841732c8

pkg/cloud/types.go
```diff
@@ -57,13 +57,28 @@ type testRunList struct {
...
+)
+
 // TestRunData holds the output from /loadtests/v4/test_runs(%s)
 type TestRunData struct {
 	TestRunId     int `json:"id"`
 	Instances     int `json:"instances"`
 	LZConfig      `json:"k8s_load_zones_config"`
 	RunStatus     cloudapi.RunStatus `json:"run_status"`
 	RuntimeConfig cloudapi.Config    `json:"k6_runtime_config"`
+	TestRunToken  string             `json:"test_run_token,omitempty"`
```
yorugac: Briefly: is this the same token that we'll be sending in `gck6_env_vars` [here](https://github.com/grafana/k6-operator/issues/742)?

internal/controller/testrun_controller.go
```diff
@@ -327,9 +327,13 @@ func (r *TestRunReconciler) reconcile(ctx context.Context, req ctrl.Request, log
...
 
-		// If this is a cloud test run in any mode, try to finalize it.
+		// If this is a non-PLZ cloud test run, try to finalize it.
+		// PLZ test runs are finalized by k6's own cloud output when the scheduler
+		// stops; the operator's FinishTestRun call (POST /v1/tests/{id}) returns
+		// 404 for PLZ runs and must be skipped to avoid getting stuck here.
 		if v1alpha1.IsTrue(k6, v1alpha1.CloudTestRun) &&
-			v1alpha1.IsFalse(k6, v1alpha1.CloudTestRunFinalized) {
+			v1alpha1.IsFalse(k6, v1alpha1.CloudTestRunFinalized) &&
+			v1alpha1.IsFalse(k6, v1alpha1.CloudPLZTestRun) {
```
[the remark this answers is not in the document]
yorugac: This is not a full review yet, but by brief look: I don't think this change should be present in this PR. Something is off...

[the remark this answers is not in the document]
vortegatorres: Reverted https://github.com/grafana/k6-operator/pull/777/changes/50cb10b00dff48719478e4af9c5d209dcc4f4231

pkg/cloud/types.go
yorugac: For brief illustration. This is what will happen very soon:
https://github.com/grafana/k6-operator/blob/779313f569ab9b4199b620b42475321c270981c7/pkg/cloud/types.go#L82-L84
There will be `K6_CLOUD_TOKEN` - which might be the same thing you're adding here. IIRC, I mentioned it before in some discussion. Could you please double-check this endpoint for duplicates? @vortegatorres

vortegatorres: I think so because this is the `TestRunToken`, which is ephemeral and only valid during test execution, and you mentioned that the mentioned in the other issue `K6_CLOUD_TOKEN` is a test-run-specific token. I would name it `TestRunToken` everywhere to avoid confusion.

pkg/cloud/types.go
```diff
@@ -95,6 +110,26 @@ func (lz *LZConfig) EnvVars() []corev1.EnvVar {
...
+	ev := []corev1.EnvVar{
+		{Name: secretSourceEnvVar, Value: "url"},
+		{Name: secretSourceURLTemplate, Value: trd.SecretsConfig.Endpoint},
+		{Name: secretSourceURLRespPath, Value: trd.SecretsConfig.ResponsePath},
+	}
+	if trd.TestRunToken != "" {
+		ev = append(ev, corev1.EnvVar{
+			Name:  secretSourceURLAuthKey,
+			Value: "Bearer " + trd.TestRunToken,
+		})
```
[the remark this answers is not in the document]
vortegatorres: I don't think this is required. @yorugac wdyt?

pkg/cloud/types.go
yorugac: Maybe. I don't have super strong opinion here (other than "avoid duplicates & conflicts" :smile:).
But this is about changing this endpoint then. It should be raised internally :eyes:

pkg/cloud/types.go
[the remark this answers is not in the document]
yorugac: Yes, since its ephemeral token, it's actually "safer" than what we had before, right? Let's leave it like this for now.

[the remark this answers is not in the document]
yorugac: One thing I'd consider is documenting this additionally. This details about ephemeral token, its TTL: it'd be nice to have them out in public.

pkg/cloud/types_test.go
```diff
@@ -60,6 +62,72 @@ func TestInspectOutput_TestNameAndProjectID(t *testing.T) {
...
+			trd: TestRunData{
+				SecretsConfig: &SecretsConfig{Endpoint: someEndpoint, ResponsePath: someRespPath},
+				TestRunToken:  someToken,
+			},
+			expected: []corev1.EnvVar{
+				{Name: "K6_SECRET_SOURCE", Value: "url"},
+				{Name: "K6_SECRET_SOURCE_URL_URL_TEMPLATE", Value: someEndpoint},
+				{Name: "K6_SECRET_SOURCE_URL_RESPONSE_PATH", Value: someRespPath},
+				{Name: "K6_SECRET_SOURCE_URL_HEADER_AUTHORIZATION", Value: "Bearer " + someToken},
+			},
```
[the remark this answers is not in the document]
vortegatorres: Fixed in https://github.com/grafana/k6-operator/pull/777/changes/2e9687bb3d302fd10351c9471bd49322af6dc588

pkg/cloud/types.go
vortegatorres: It is not the same token: https://github.com/grafana/k6-cloud-general-api/blob/077317907c1487ee9a98b303ab08e7910f96cfc1/k6/serializers/load_test_run.py#L271

yorugac: What's the difference between them then?

pkg/cloud/types.go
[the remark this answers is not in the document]
vortegatorres: The Test Run Token is documented here: https://github.com/grafana/k6-internal-docs/blob/3c6441427b6dceba0391dcc0b4c52153c714ddb3/docs/backend/executing-load-tests/secrets-configuration.md#test-run-token

pkg/cloud/types.go
vortegatorres: - `test_run_token` is a short-lived, test-run-scoped token, auto-generated when the run is created and auto-invalidated when it ends. Documented here: https://github.com/grafana/k6-internal-docs/blob/3c6441427b6dceba0391dcc0b4c52153c714ddb3/docs/backend/executing-load-tests/secrets-configuration.md#test-run-token
 - `K6_CLOUD_TOKEN` in `gck6_env_vars` is the user's API token (or org service token), the same long-lived token sent to the test coordinator as `TC_USER_ACCESS_TOKEN`. It's used for general cloud operations (metrics ingestion, log streaming, traces, etc.).                

They're different tokens with different scopes, lifetimes, and auth backends. `test_run_token` is the correct one for secrets decryption because it enforces least-privilege: a compromised runner can only access its own secrets, not the user's entire account.

yorugac: AFAIK, the token in `gck6_env_vars` is also not supposed to live longer than the test run though.

But the main diff is in scope then :thinking: Can secrets be read by that other one ("user's API token")?

vortegatorres: No, secrets can be read just using the `test_run_token`.

yorugac: OK, so the token in this PR is fully special for secrets reading specifically.

pkg/cloud/types.go
```diff
@@ -57,13 +57,28 @@ type testRunList struct {
 	} `json:"object"`
 }
 
+// SecretsConfig holds the secrets configuration returned by k6 Cloud.
+type SecretsConfig struct {
+	Endpoint     string `json:"endpoint"`
+	ResponsePath string `json:"response_path"`
```
yorugac, line 63: I assume these are always non-empty.

pkg/cloud/types_test.go
```diff
@@ -60,6 +62,72 @@ func TestInspectOutput_TestNameAndProjectID(t *testing.T) {
...
+		trd      TestRunData
+		expected []corev1.EnvVar
+	}{
+		{
+			name:     "nil secrets config returns nil",
+			trd:      TestRunData{},
+			expected: nil,
+		},
+		{
+			name: "secrets config without token omits auth header",
```
yorugac: Is this case valid in practice?

pkg/cloud/types_test.go
```diff
@@ -60,6 +62,72 @@ func TestInspectOutput_TestNameAndProjectID(t *testing.T) {
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			got := tt.trd.SecretsEnvVars()
+			if tt.expected == nil {
+				if got != nil {
+					t.Errorf("SecretsEnvVars() = %+v, want nil", got)
+				}
+				return
+			}
```
yorugac: Isn't this already covered by the check on line 119?

pkg/cloud/types_test.go
```diff
@@ -60,6 +62,72 @@ func TestInspectOutput_TestNameAndProjectID(t *testing.T) {
...
+	t.Parallel()
+
+	someEndpoint := "https://api.k6.io/provisioning/v1/test_runs/42/decrypt_secret?name={key}"
+	someRespPath := "plaintext"
+	someToken := "abc123"
+
+	tests := []struct {
+		name     string
+		trd      TestRunData
+		expected []corev1.EnvVar
```
yorugac:
```suggestion
		expected []corev1.EnvVar // order of env vars is important
```

pkg/plz/worker_test.go
```diff
@@ -227,6 +231,29 @@ func Test_complete_correctDefinitionOfTestRun(t *testing.T) {
...
+		corev1.EnvVar{Name: "K6_SECRET_SOURCE_URL_HEADER_AUTHORIZATION", Value: "Bearer " + someTestRunToken},
+	)
+
+	secretsWithoutTokenTestRun = cloudFieldsTestRun
+	secretsWithoutTokenTestRun.Spec.Runner.Env = append(
+		append([]corev1.EnvVar{}, cloudFieldsTestRun.Spec.Runner.Env...),
+		corev1.EnvVar{Name: "K6_SECRET_SOURCE", Value: "url"},
+		corev1.EnvVar{Name: "K6_SECRET_SOURCE_URL_URL_TEMPLATE", Value: someSecretsConfig.Endpoint},
+		corev1.EnvVar{Name: "K6_SECRET_SOURCE_URL_RESPONSE_PATH", Value: someSecretsConfig.ResponsePath},
+	)
```
yorugac: The same question as above: is this case valid in practice?

pkg/cloud/types.go
yorugac:
```suggestion
    // token for read-only access to secrets
	SecretsToken  string             `json:"test_run_token,omitempty"`
```
I'd like to rename it to reflect the reality. 
IMO, API should also be using something different to `test_run_token` because its confusing. But let's have some clarity at least at the code level.

pkg/cloud/types.go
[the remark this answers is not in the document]
yorugac: Can we move these couple of sections to public docs? Specifically the "Test Run Token" (-> "Test run token for secrets") and "Security Benefits" parts.

Now that I ran some tests, it is a bit disconcerting that we have one more token lying around and accessible as env var... The token to get secrets is not actually secret itself. So yeah, let's be as clear as possible about what it is and its scope.

yorugac (changes requested):
Hi @vortegatorres,
Thanks for all the updates and clarifications! I have run some basic tests with this:
```js
import tempo from 'https://jslib.k6.io/http-instrumentation-tempo/1.0.0/index.js';
import { sleep } from "k6";
import secrets from "k6/secrets";

tempo.instrumentHTTP({
  propagator: "w3c",
});

export const options = {
  cloud: {
    distribution: {
      "xxxx": { loadZone: "xxxx", percent: 100 },
    },
    projectID: xxxx,
  },
  thresholds: {
    http_req_duration: ['p(95)<2000'], // 95% of requests should be below 200ms
  },
  scenarios: {
    contacts: {
      executor: 'per-vu-iterations',
      vus: 10,
      iterations: 200,
      maxDuration: '60s',
    },
  },
};

export default async function main() {
  const mystery = await secrets.get("mystery");
  console.log("Secret is", mystery);

  console.log("Env var token ", __ENV.K6_SECRET_SOURCE_URL_HEADER_AUTHORIZATION);

  sleep(1);
}
```
This seems to work 🙌 
It's a bit concerning to have this token readable but since it's so limited in scope, I guess we can live with it, at least for some time. 

Otherwise, I have some small suggestions for the code: please see the comments. Also, I wonder about unit tests for "no token": does GCk6 allow such a thing? If yes, why?

pkg/cloud/types.go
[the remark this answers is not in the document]
vortegatorres: I don’t think moving this to the public documentation is a good idea. The test run token is an internal token passed between certain k6 components. Users should never need to handle it directly. Since it is also part of the k6 Cloud secrets feature, I believe it is better to keep public information about it to a minimum. While someone determined could still infer some of this from the public k6-operator code, limiting what we document publicly still seems like the safer approach. @facundobatista do you agree?

pkg/cloud/types.go
vortegatorres: Sorry, but I disagree. Right now, it is only used for secrets, but in the future, we plan to use it for sending metrics and potentially other purposes as well. That is why I did not include it in the secrets configuration section.

pkg/cloud/types_test.go
vortegatorres: Yes thanks, fixed in https://github.com/grafana/k6-operator/pull/777/changes/de7975ccedbef0eedf2c623c4ed6dc7629564688

pkg/cloud/types_test.go
vortegatorres: You're right to question it. `test_run_token` is always generated when a test run is created, so in practice if `secrets_config` is present the token will always be present too. Removed both test cases in https://github.com/grafana/k6-operator/pull/777/changes/5a72610176d70ae5df5b5bd6a271d0e00d214bac

pkg/cloud/types.go
vortegatorres: Yes, when `secrets_config` is returned by the API, both `endpoint` and `response_path` will always be populated.

pkg/plz/worker_test.go
vortegatorres: You're right to question it. test_run_token is always generated when a test run is created, so in practice if secrets_config is present the token will always be present too. Removed both test cases in https://github.com/grafana/k6-operator/commit/5a72610176d70ae5df5b5bd6a271d0e00d214bac

pkg/cloud/types.go
yorugac: That's what I've been asking about before :sweat_smile: Here:
https://github.com/grafana/k6-operator/pull/777/changes#r3071576522
You said these two tokens have clear difference in usage. And metrics are supposed to be done by the second one - from another issue / PR. So this token, in this PR, is only for secrets for foreseeable (6m+) future. That's how I read it at least. Is this not so?

vortegatorres: It’s only for secrets right now, but we may use it for metrics in the future as well. I can rename the token, keeping in mind that we might eventually want to revert to the original name, `TestRunToken`, to reflect that it is a token used only for a test and exists only for the duration of that test. In my head was better to call it `TestRunToken` everywhere but I can change it.

vortegatorres: @yorugac I have renamed it to `SecretsToken`, we can renamed again later in https://github.com/grafana/k6-operator/pull/777/changes/e3e835da20d52696932fce3b0ce7a247f879a1e0

yorugac: :thinking: For my understanding, is it something that you've considered specifically for v6? I.e. expanding the scope of this token.
Also, could we just unite it all under one token now?

vortegatorres: It’s more of an idea than something concrete, we don’t know yet.

yorugac (approved):
Thanks for all the updates, @vortegatorres!

pkg/cloud/types.go
yorugac: Thanks!

Totally agree that we can always rename it later, if it changes its role. TBH, long-term, the main toll of backwards compatibility for this API would be on the GCk6. 
I'll share info about this token on the squad sync too: maybe there's a path forward to simplify these parallel tokens.

PR #781, opened 2026-04-24
Change job_name tag for testrun_name

### LINKED ISSUE #609: Change the default tag `job_name` to `testrun_name`
### Feature Description

Currently, we have two k6 tags added to all tests runs:
    - `instance_id` - an index of the k6 segment taken (1, ... n)
    - `job_name` - a `Job`'s name which is basically a `TestRun` name with the same index (`my-testrun-name-%d`)

These two are duplicating information as they are both using the segment's index. As a consequence, both will have the same impact on cardinality of metrics. 

To improve this, we could remove `job_name` tag and add `testrun_name` tag instead.

If there are drawbacks to this proposal, please let us know in the comments.

_No response_

federicotdn:
Fixes https://github.com/grafana/k6-operator/issues/609.

Remove the `job_name` tag in test run Jobs, and add a new `testrun_name` tag that only contains the name without instance ID.


yorugac (approved):
🙌 Thank you, @federicotdn!

Issue #627, opened 2025-08-06
Add SchedulerName to starter, runner and initializer pod


sharathjag:
### Feature Description

Add SchedulerName to starter, runner and initializer pod. 
Currently SchedulerName is not an attribute in Pod spec: https://github.com/grafana/k6-operator/blob/13d5108137ee70056869eb3cb47b971e6369e48a/api/v1alpha1/testrun_types.go#L34

sharathjag:
Happy to contribute.

sharathjag:
https://github.com/grafana/k6-operator/compare/main...sharathjag:k6-operator:issue-627?expand=1
Happy to make necessary changes. Please let me know how to proceed. 
🙏

sharathjag:
@yorugac 
Thank you.

yorugac:
Hi @sharathjag, could you please describe your use case in greater details? Does your cluster actually have multiple schedulers or do you expect something different from this value? The branch you linked doesn't do anything with scheduler name so I started to wonder if I understand correctly 🤔

sharathjag:
Thanks @yorugac 

> Does your cluster actually have multiple schedulers or do you expect something different from this value?

Yes, we will have multiple schedulers supported on the cluster. For our use-case, we will be using Volcano scheduler benefitting from its queues and quota management.  


> The branch you linked doesn't do anything with scheduler name

Likely my lack of understanding of the code. I was intending to add schedulerName to the [TestRun](https://github.com/grafana/k6-operator/blob/main/api/v1alpha1/testrun_types.go#L83) spec and update Runner Spec in https://github.com/grafana/k6-operator/blob/ad6c08457bdffaf2f68202bad5f10cc13025d192/pkg/resources/jobs/runner.go#L177 

If that is not what is needed, happy to coordinate on whatever is necessary

yorugac:
Hi @sharathjag, thanks for the details.


I'm not sure how often people hit this problem (your issue is literally the first time a scheduler came up), so I'm curious about community feedback.
For example, it is likely possible to assign a custom scheduler to all pods in the given group with the mutating webhook. There could be solutions that allow one to do that quickly, i.e. an already implemented webhook. The quick search points me to the Gatekeeper, but maybe there are others.

> I was intending to add schedulerName to the [TestRun](https://github.com/grafana/k6-operator/blob/main/api/v1alpha1/testrun_types.go#L83) spec and update Runner Spec in

[...]

PR #783, opened 2026-04-26
Adding support to specify SchedularName (optional) in Initializer/Starter/Runner/Stopper Specs/Jobs

The-Flash-Routine:
**Issue**:
https://github.com/grafana/k6-operator/issues/627

**PR Changes**:
1. Adding support to specify SchedularName
2. Added test cases for both scenario in initializer/Starter/Runner/Stopper
3. Added ".go-version" in gitignore as I use "goenv" CLI tool to maintain multiple versions of GO. Other people might use that too

**Thoughts**:
I understand that using a custom scheduler is a niche use-case, but we should still support all specifications of a Kubernetes Pod when we create a Job

**Additional**:
I've also raised this improvement issue (easy change)
https://github.com/grafana/k6-operator/issues/784

---

> [!NOTE]
> **Low Risk**
> Adds an optional `schedulerName` field that influences how initializer/starter/runner/stopper Jobs are scheduled; risk is low but could affect deployments that rely on a non-default scheduler configuration.
> 
> **Overview**
> Adds optional `spec.*.schedulerName` support to `TestRun` so initializer, runner, and starter Pods set `PodSpec.SchedulerName` (defaulting to Kubernetes `default-scheduler`), with stopper inheriting it via the starter template.
> 
> Updates the CRD schema and generated docs to expose the new field, and refactors job unit tests into table-driven cases covering both default and custom scheduler names.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b85276cd24c7b2ef6f3a1409fb111ab1a47bb6a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ testbin
 *.env
 
 .devspace/
+
+.go-version
--- a/.gitignore
+++ b/.gitignore
@@ -33,5 +33,3 @@ testbin
 *.env
 
 .devspace/
-
-.go-version
--- a/pkg/resources/jobs/initializer.go
+++ b/pkg/resources/jobs/initializer.go
@@ -24,6 +24,7 @@ func NewInitializerJob(k6 *v1alpha1.TestRun, argLine string) (*batchv1.Job, erro
 		serviceAccountName           = "default"
 		automountServiceAccountToken = true
 		ports                        = append([]corev1.ContainerPort{{ContainerPort: 6565}}, k6.GetSpec().Ports...)
+		schedulerName                = "default-scheduler"
 	)
 
 	if k6.GetSpec().Initializer == nil {
@@ -94,6 +95,10 @@ func NewInitializerJob(k6 *v1alpha1.TestRun, argLine string) (*batchv1.Job, erro
 	volumeMounts := script.VolumeMount()
 	volumeMounts = append(volumeMounts, k6.GetSpec().Initializer.VolumeMounts...)
 
+	if k6.GetSpec().Initializer.SchedulerName != "" {
+		schedulerName = k6.GetSpec().Initializer.SchedulerName
+	}
+
 	var zero32 int32
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
@@ -118,6 +123,7 @@ func NewInitializerJob(k6 *v1alpha1.TestRun, argLine string) (*batchv1.Job, erro
 					TopologySpreadConstraints:    k6.GetSpec().Initializer.TopologySpreadConstraints,
 					SecurityContext:              &k6.GetSpec().Initializer.SecurityContext,
 					RestartPolicy:                corev1.RestartPolicyNever,
+					SchedulerName:                schedulerName,
 					ImagePullSecrets:             k6.GetSpec().Initializer.ImagePullSecrets,
 					InitContainers:               getInitContainers(k6.GetSpec().Initializer, script),
 					Containers: []corev1.Container{
--- a/pkg/resources/jobs/initializer.go
+++ b/pkg/resources/jobs/initializer.go
@@ -24,7 +24,7 @@ func NewInitializerJob(k6 *v1alpha1.TestRun, argLine string) (*batchv1.Job, erro
 		serviceAccountName           = "default"
 		automountServiceAccountToken = true
 		ports                        = append([]corev1.ContainerPort{{ContainerPort: 6565}}, k6.GetSpec().Ports...)
-		schedulerName                = "default-scheduler"
+		schedulerName                = corev1.DefaultSchedulerName
 	)
 
 	if k6.GetSpec().Initializer == nil {
--- a/pkg/resources/jobs/initializer_test.go
+++ b/pkg/resources/jobs/initializer_test.go
@@ -51,6 +51,7 @@ func Test_NewInitializerJob(t *testing.T) {
 				Spec: corev1.PodSpec{
 					AutomountServiceAccountToken: &automountServiceAccountToken,
 					ServiceAccountName:           "default",
+					SchedulerName:                "default-scheduler",
 					Affinity:                     nil,
 					NodeSelector:                 nil,
 					Tolerations:                  nil,
@@ -252,5 +253,128 @@ func Test_InitializerEnvVarFlags(t *testing.T) {
 				}
 			}
 		})
+	}
+}
+
+func Test_NewInitializerJobWithSchedulerName(t *testing.T) {
+	script := &types.Script{
+		Name:     "test",
+		Filename: "test.js",
+		Type:     "ConfigMap",
+	}
+
+	automountServiceAccountToken := true
+	zero := int32(0)
+
+	expectedOutcome := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-initializer",
+			Namespace: "test",
+			Labels: map[string]string{
+				"app":    "k6",
+				"k6_cr":  "test",
+				"label1": "awesome",
+			},
+			Annotations: map[string]string{
+				"awesomeAnnotation": "dope",
+			},
+		},
+		Spec: batchv1.JobSpec{
+			BackoffLimit: &zero,
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"app":    "k6",
+						"k6_cr":  "test",
+						"label1": "awesome",
+					},
+					Annotations: map[string]string{
+						"awesomeAnnotation": "dope",
+					},
+				},
+				Spec: corev1.PodSpec{
+					AutomountServiceAccountToken: &automountServiceAccountToken,
+					ServiceAccountName:           "default",
+					SchedulerName:                "custom-scheduler",
+					Affinity:                     nil,
+					NodeSelector:                 nil,
+					Tolerations:                  nil,
+					TopologySpreadConstraints:    nil,
+					RestartPolicy:                corev1.RestartPolicyNever,
+					SecurityContext:              &corev1.PodSecurityContext{},
+					Containers: []corev1.Container{
+						{
+							Image:           "grafana/k6:latest",
+							ImagePullPolicy: "",
+							Name:            "k6",
+							Command: []string{
+								"sh", "-c",
+								"mkdir -p $(dirname /tmp/test.js.archived.tar) && k6 archive /test/test.js -O /tmp/test.js.archived.tar --out cloud 2> /tmp/k6logs && k6 inspect --execution-requirements /tmp/test.js.archived.tar 2> /tmp/k6logs ; ! cat /tmp/k6logs | grep 'level.*error'",
+							},
+							Env: []corev1.EnvVar{},
+							EnvFrom: []corev1.EnvFromSource{
+								{
+									ConfigMapRef: &corev1.ConfigMapEnvSource{
+										LocalObjectReference: corev1.LocalObjectReference{
+											Name: "env",
+										},
+									},
+								},
+							},
+							Resources:       corev1.ResourceRequirements{},
+							VolumeMounts:    script.VolumeMount(),
+							Ports:           []corev1.ContainerPort{{ContainerPort: 6565}},
+							SecurityContext: &corev1.SecurityContext{},
+						},
+					},
+					Volumes: script.Volume(),
+				},
+			},
+		},
+	}
+
+	k6 := &v1alpha1.TestRun{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test",
+			Namespace: "test",
+		},
+		Spec: v1alpha1.TestRunSpec{
+			Script: v1alpha1.K6Script{
+				ConfigMap: v1alpha1.K6Configmap{
+					Name: "test",
+					File: "test.js",
+				},
+			},
+			Arguments: "--out cloud",
+			Initializer: &v1alpha1.Pod{
+				Metadata: v1alpha1.PodMetadata{
+					Labels: map[string]string{
+						"label1": "awesome",
+					},
+					Annotations: map[string]string{
+						"awesomeAnnotation": "dope",
+					},
+				},
+				EnvFrom: []corev1.EnvFromSource{
+					{
+						ConfigMapRef: &corev1.ConfigMapEnvSource{
+							LocalObjectReference: corev1.LocalObjectReference{
+								Name: "env",
+							},
+						},
+					},
+				},
+				SchedulerName: "custom-scheduler",
+			},
+		},
+	}
+
+	job, err := NewInitializerJob(k6, "--out cloud")
+	if err != nil {
+		t.Errorf("NewInitializerJob errored, got: %v", err)
+	}
+
+	if diff := deep.Equal(job, expectedOutcome); diff != nil {
+		t.Error(diff)
 	}
 }
--- a/pkg/resources/jobs/initializer_test.go
+++ b/pkg/resources/jobs/initializer_test.go
@@ -12,7 +12,44 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func Test_NewInitializerJob(t *testing.T) {
+func defaultTestRunForInitializer() *v1alpha1.TestRun {
+	return &v1alpha1.TestRun{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test",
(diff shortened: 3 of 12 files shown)


The-Flash-Routine:
Hi @yorugac 
Can you please help me enable the workflow checks

The-Flash-Routine:
Hi @yorugac 
Can you please help me enable workflows

yorugac:
Hi @The-Flash-Routine, sorry for the delay, this week is pretty overwhelming! I'll be able to look at this next week. (CI unlocked)
Thank you!

The-Flash-Routine:
Thanks @yorugac :)

The-Flash-Routine:
Hi @yorugac 
Can you please help me with the PR

pkg/resources/jobs/initializer.go
```diff
@@ -24,6 +24,7 @@ func NewInitializerJob(k6 *v1alpha1.TestRun, argLine string) (*batchv1.Job, erro
 		serviceAccountName           = "default"
 		automountServiceAccountToken = true
 		ports                        = append([]corev1.ContainerPort{{ContainerPort: 6565}}, k6.GetSpec().Ports...)
+		schedulerName                = "default-scheduler"
```
yorugac: It's preferable to use the value from the lib instead of a hard-coding one. So probably this one:
https://pkg.go.dev/k8s.io/api/core/v1#DefaultSchedulerName

.gitignore
```diff
@@ -32,4 +32,6 @@ testbin
 
 *.env
 
-.devspace/
\ No newline at end of file
+.devspace/
+
+.go-version
```
yorugac: Please remove this change. You mentioned goenv in description: certainly feel free to use it, but it's not an official Go tooling and k6-operator neither requires nor recommends it. So it's more fit for a personal gitignore.

pkg/resources/jobs/runner.go
```diff
@@ -80,8 +80,9 @@ func NewRunnerJob(k6 *v1alpha1.TestRun, index int, tokenInfo *cloud.TokenInfo) (
 	command = script.UpdateCommand(command)
 
 	var (
-		zero   int64 = 0
-		zero32 int32 = 0
+		zero          int64 = 0
+		zero32        int32 = 0
+		schedulerName       = "default-scheduler"
```
yorugac: Same as above.

pkg/resources/jobs/starter.go
```diff
@@ -15,6 +15,10 @@ import (
 // NewStarterJob builds a template used for creating a starter job
 func NewStarterJob(k6 *v1alpha1.TestRun, hostname []string) *batchv1.Job {
 
+	var (
+		schedulerName = "default-scheduler"
```
yorugac: Same as above.

pkg/resources/jobs/initializer_test.go
```diff
@@ -254,3 +255,126 @@ func Test_InitializerEnvVarFlags(t *testing.T) {
 		})
 	}
 }
+
+func Test_NewInitializerJobWithSchedulerName(t *testing.T) {
```
yorugac: This is a very large test 😅 It'd be nice to rewrite these as `runner_test.go` was done. I.e. this should be part of `Test_NewInitializerJob` function as a separate test case. Would that be possible?

pkg/resources/jobs/starter_test.go
```diff
@@ -282,3 +284,98 @@ func TestNewStarterJobCustomResources(t *testing.T) {
 		t.Errorf("custom resources not applied: %v", diff)
 	}
 }
+
+func TestNewStarterJobWithSchedulerName(t *testing.T) {
```
yorugac: Similar situation as with initializer tests. It'd be nice to rewrite this as test case instead of function.

pkg/resources/jobs/stopper_test.go
```diff
@@ -189,3 +191,87 @@ func TestNewStopJobIstio(t *testing.T) {
 		t.Error(diff)
 	}
 }
+
+func TestNewStopperJobWithSchedulerName(t *testing.T) {
```
yorugac: Similar situation as with initializer tests. It'd be nice to rewrite this as test case instead of function.

yorugac (changes requested):
Hi @The-Flash-Routine, thanks again for the PR!

I've left some comments: half of them should be trivial change requests, please see.

But there are 3 comments for unit test refactoring: these are a bit more involved, so no pressure to do this. I.e. f you're not up to it, then we can probably omit corresponding unit tests. But if you'd like to, feel free to refactor and keep the tests too. (I'd personally prefer refactoring, just don't want to ask too much given this addition is pretty straight-forward 😄)

pkg/resources/jobs/initializer.go
The-Flash-Routine: Good suggestion @yorugac . Thanks :)
I will make a change for "SchedulerName".

I see the same suggestion working for other existing things like "ServiceAccount", "AutomountServiceAccountToken" as well.
Should I make change for them too ?

.gitignore
The-Flash-Routine: @yorugac but it's just an addition into gitignore file so that I (or anyone) don't accidentally checkin that file.

It's even used in big projects like git repo for Kubernetes.

Can you please help me give it another consideration.

pkg/resources/jobs/runner.go
The-Flash-Routine: Similar question as above, I will pick default of SchedulerName value from library, but shall I make that change for other existing stuff too ? Like "SchedulerName" etc

pkg/resources/jobs/starter.go
The-Flash-Routine: Similar question as above

pkg/resources/jobs/initializer_test.go
The-Flash-Routine: @yorugac it is very well possible and I can do it as well, but I do have a difference of opinion on a few things:

1.
In general I've faced a lot of issues in past when tests are interlinked with each other.
For example, I see that tests in **runner_test.go** uses common functions like **defaultTestRun()** (etc)
These things create a problem that change in one such linked function breaks other test cases.

I'm a strong believer of tests being independent of each other. Not linked to other tests in any way.
The **Input** and **Expected_Output** for each test should be fully defined inside that test itself.

Now I agree it increases size of testing file, but that does give a lot of structure and clarity in code such that

a.
You can give test proper names.
So when someone makes a change and let's say something breaks, then it very clearly shows that which test broke:
Example:
**Test_SchedularName** or **Test_Affinity**

b.
Let's say if someone makes an addition, they know via test names, that which tests they should be looking for specifically.
It makes developing very easy in that manner

2.
Honestly, I was even thinking of raising an issue to refactor tests in runner_test.go

==

@yorugac can you please have a look at above reasoning, if you still think I should make changes, I will do that.

pkg/resources/jobs/starter_test.go
The-Flash-Routine: similar question as above for initializer_test.go

pkg/resources/jobs/stopper_test.go
The-Flash-Routine: similar question as above for initializer_test.go

pkg/resources/jobs/initializer.go
The-Flash-Routine: @yorugac I've made the change for this. Now we are using the default value from kubernetes package.

pkg/resources/jobs/runner.go
The-Flash-Routine: @yorugac I've made the change for this. Now we are using the default value from kubernetes package.

pkg/resources/jobs/starter.go
The-Flash-Routine: @yorugac I've made the change for this. Now we are using the default value from kubernetes package.

.gitignore
yorugac: Kubernetes has go-version explicitly, not in gitignore... So they use it as part of their setup and CI - explicitly so. 
k6-operator has only recently switched to `go toolchain`, which Kubernetes is not using yet, but [might switch](https://github.com/kubernetes/kubernetes/issues/130799) as well.

But either way, this is about gitignore - not about whether we should use goenv. TBH, I have quite a lot of entries in a personal gitignore that I'm not pushing into the project 😅 If we add this one, what else should we add? Is there an end to it?

Just in case, I don't think this topic can ever have a "universal" answer. This is mostly a "how one prefers to arrange things in the project". My preference is to have a clear separation between what can be generated by the project (part of `.gitignore`) and what artifacts are created by me or my own tooling (part of personal gitignore). I don't see much of a point in having a large pot of everything in a joint `.gitignore` because everyone has their own workflows.

pkg/resources/jobs/initializer_test.go
yorugac: @The-Flash-Routine, Yes, I'm familiar with these caveats around unit tests as well. Your argument definitely has merit, but I think it should be evaluated on a case by case basis, i.e. it's not a general rule. We're not talking about any kind of complex logic here, but specifically _unit tests for generated yaml specs_.

Having lived with 3 different types of testing systems for yaml specs in this very project, my current preference is what was last done in `runner_test.go`: with `setupTestRun` function. 

> You can give test proper names.

Test cases also have unique names. I'd even argue they are better since they have a more relaxed format than names of functions.

> It makes developing very easy in that manner

It makes developing easier, and it makes long-term maintenance harder 😂 

These tests around TestRun specs are pretty straight-forward: `defaultTestRun` had minimal changes through years and there's not that much to keep track of or get confused about in the test itself. It's a simple 1:1 transformation of data but with many small conditionals that should be checked.

pkg/resources/jobs/initializer.go
yorugac: > I see the same suggestion working for other existing things like "ServiceAccount", 
> "AutomountServiceAccountToken" as well.
> Should I make change for them too ?

WDYM? By my quick search, I didn't see defaults for those as exported vars... Unless I missed them?

The-Flash-Routine: @yorugac 
Well it's a bit weird, it's here.
https://pkg.go.dev/k8s.io/kubernetes@v1.36.0/plugin/pkg/admission/serviceaccount

Well technically if they would put all defaults in one place, it would be much better.

.gitignore
The-Flash-Routine: @yorugac Fair enough ! I'll remove this from PR and follow your footsteps (keep it in personal gitignore)

pkg/resources/jobs/initializer_test.go
The-Flash-Routine: @yorugac let's just agree to disagree on this topic.

But I also respect your experience and being the sole maintainer of this project you are doing pretty amazing job, I'm impressed, so I'll make the change and make tests look like **runner_tests.go**

pkg/resources/jobs/stopper_test.go
The-Flash-Routine: Conversation on above comment, will make tests look like **runner_tests.go**

pkg/resources/jobs/starter_test.go
The-Flash-Routine: Conversation on above comment, will make tests look like **runner_tests.go**

pkg/resources/jobs/initializer.go
yorugac: Oh I see, it's over there... Well no, just having a fancy const referenced here is not worth it importing this package, IMO.
Let's keep it as is. Thanks for looking!

pkg/resources/jobs/initializer_test.go
yorugac: :sweat_smile: Thanks. 
TBH, in another case, I'd agree with you. Just not in the case of these walls of text produced for yaml specs here!

pkg/resources/jobs/initializer.go
The-Flash-Routine: @yorugac okay !

.gitignore
The-Flash-Routine: @yorugac Removed from .gitignore !

pkg/resources/jobs/initializer_test.go
The-Flash-Routine: @yorugac 
Please find the following are changes in this file:
1.
I've refactored the original **Test_NewInitializerJob** to match the **runner_test.go** testing style
2.
I've removed my original **Test_NewInitializerJobWithSchedulerName** and instead added to the struct list test in **Test_NewInitializerJob**
3.
[Important]
The reason I didn't refactor the **Test_InitializerEnvVarFlags** (and didn't add to the struct list test) is because I see in that, we are testing "Runner" stuff, which is bit odd that in "initializer_test.go" we have tests for Runner.
That is why I've left it as it is.

pkg/resources/jobs/starter_test.go
The-Flash-Routine: @yorugac 
Please find following are changes in this file:
1.
I've refactored the original **Test_NewStarterJob** to match the **runner_test.go** testing style
2.
I've removed my original **Test_NewStarterJobWithSchedulerName** and instead added to the struct list test in **Test_NewStarterJob**
3a.
I've refactored **TestNewStarterJobCustomResources** and added it to struct list test in **Test_NewStarterJob**
3b.
In these I've also made an improvement to NOT use these:
resource.NewMilliQuantity(50, resource.DecimalSI)
but instead use these:
resource.MustParse("100m")

pkg/resources/jobs/stopper_test.go
The-Flash-Routine: @yorugac 
Please find the following changes in this file:

1.
I've refactored the original **Test_NewStopperJob** to match the **runner_test.go** testing style
2.
I've removed my original **Test_NewStopperJobWithSchedulerName** and instead added to the struct list test in **Test_NewStopperJob**
3.
Stopper is using the same API specs of Starter.
Will raise an issue for this

pkg/resources/jobs/starter_test.go
```diff
@@ -70,39 +100,86 @@ func TestNewStarterJob(t *testing.T) {
...
+				j.Spec.Template.Spec.SchedulerName = "custom-scheduler"
+			},
+		},
+		{
+			name:     "custom resources",
+			hostname: []string{"testing"},
+			setupTestRun: func(k6 *v1alpha1.TestRun) {
+				k6.Spec.Starter.Resources = corev1.ResourceRequirements{
+					Limits: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("100m"),
```
yorugac, line 133: This can now panic, IIRC. Since it's unit tests, we can likely live with that, but worth to remember.

yorugac (approved):
Thank you for the changes and esp. the tests, @The-Flash-Routine 🙌

pkg/resources/jobs/stopper_test.go
yorugac: > Stopper is using the same API specs of Starter. Will raise an issue for this

Not sure I follow. What is the issue with that exactly?

pkg/resources/jobs/initializer_test.go
yorugac: Yes, `Test_InitializerEnvVarFlags` should be separate :+1: And it's not testing runner, it's testing initializer logic only. But initializer spec is equal to runner's unless set explicitly, that's why it ended up that way.

The-Flash-Routine: Ah. Okay ! Thanks @yorugac

pkg/resources/jobs/stopper_test.go
The-Flash-Routine: @yorugac 
um... was just hoping that Stopper should have it's own function/implementation and not use Starter's implementation.
Ignore this please, if it was done for a reason.

The-Flash-Routine:
Thanks @yorugac for help :) 
Really appreciate it ! :)

pkg/resources/jobs/starter_test.go
The-Flash-Routine: Ah. Okay. Thanks.

pkg/resources/jobs/stopper_test.go
yorugac: Ah I see. Yes, it was intentional: just an attempt to re-use. There's a [comment](https://github.com/grafana/k6-operator/blob/main/pkg/resources/jobs/stopper.go#L14) on that.

PR #788, opened 2026-04-29
Bug fix release: PLZ CRD in Helm chart must follow patch validations

yorugac:
With addition of `.podTemplate` in PLZ CRD in #726, a bug in Helm chart was inadvertently introduced: skipped validation patch resulted in a full definition being rendered instead of a limited one. Fixing it now and making a bug fix Helm release.


charts/k6-operator/templates/crds/plz.yaml
```diff
@@ -20,3929 +19,195 @@ spec:
...
-                  x-kubernetes-map-type: atomic
-                type: array
-              nodeSelector:
-                additionalProperties:
-                  type: string
-                type: object
-              podTemplate:
-                properties:
-                  metadata:
-                    type: object
```
yorugac, line 87: This was being silently pruned.

charts/k6-operator/templates/crds/plz.yaml
```diff
@@ -20,3929 +19,195 @@ spec:
...
+                  additionalProperties:
+                    type: string
+                  type: object
+                podTemplate:
+                  properties:
+                    metadata:
+                      properties:
+                        annotations:
+                          additionalProperties:
+                            type: string
```
yorugac, line 89: Explicit definitions for `podTemplate.metadata` shouldn't be pruned anymore.

charts/k6-operator/templates/crds/plz.yaml
```diff
@@ -20,3929 +19,195 @@ spec:
...
-                              type: object
-                          required:
-                          - name
-                          type: object
-                        type: array
-                        x-kubernetes-list-map-keys:
-                        - name
-                        x-kubernetes-list-type: map
                     required:
-                    - containers
```
yorugac, line 3856: And this 😞 Also polluting the validation.

yorugac:
Patched Helm is released as [4.4.1](https://github.com/grafana/helm-charts/releases/tag/helm-k6-operator-4.4.1). This release fixes a bug in Makefile tooling and, as a consequence, a bug with PLZ CRD definition in the Helm chart.

Due to this bug, when k6-operator Helm chart was installed (versions 4.3.0 - 4.4.0), the installed PLZ CRD was incorrect. It drifted away from the kustomize and `bundle.yaml` definitions, because validation patch for `spec.podTemplate` field was not applied specifically for Helm chart. As a consequence:
- `spec.podTemplate.metadata` in PLZ could be silently dropped due to [CRD field pruning](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning)
  - Potential impact: it was possible to set those fields in PLZ but they were not persisted and effectively ignored.
- `spec.podTemplate.metadata` in PLZ could also error out during creation on clusters with strict validation. See [Kubernetes docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#field-validation).
  - Potential impact: impossible to create a PLZ.
- PLZ CRD suggested that all kinds of fields are supported in `spec.podTemplate` which is explicitly not the case (release notes [1.3.0](https://github.com/grafana/k6-operator/releases/tag/v1.3.0) and [1.4.0](https://github.com/grafana/k6-operator/releases/tag/v1.4.0)).
  - Potential impact: it was possible to set unsupported fields, but they were ignored by the k6-operator.
- `spec.podTemplate.spec.containers` was marked as a required in that invalid Helm schema, and if there was a pre-existing PLZ (without containers), it became impossible to delete it due to validation error on finalizer.
  - Potential impact: impossible to delete a PLZ.

Impact is marked as potential, as the exact details depend on specifics of versions, cluster, PLZ format, what was already installed VS what was being upgraded, etc.

Issue #787, opened 2026-04-29
Add Cloud secrets support in cloud output mode


yorugac:
### Feature Description

Since https://github.com/grafana/k6-operator/issues/747 (v1.4.0) k6-operator supports Cloud Secrets in PLZ mode, but not in cloud output. For feature parity, we should expand that support.

### Suggested Solution (optional)

k6 CLI has a reference to an extended API:
https://github.com/grafana/k6/blob/master/cloudapi/api.go#L58
This approach can likely be re-used.

It also might make sense to do this addition as part of a switch to new API v6 endpoint.

As additional context: watch out for https://github.com/grafana/k6/pull/5875 PR as it simplifies args that need to be passed to k6 CLI for secrets.

### Already existing or connected issues / PRs (optional)

https://github.com/grafana/k6-operator/issues/747

PR #789, opened 2026-04-29
[feat] Get secrets configuration from `/v1/test` in cloud output mode to support `Secrets`

vortegatorres:
https://github.com/grafana/k6-operator/issues/787

When a TestRun uses `--out cloud`, the operator creates the cloud test run via POST `/v1/tests`, and now it gets the secrets configuration from `/v1/test`, and supports secrets.

Tested here: https://admin-prod-us-east-0.grafana.net/directory-k6-cloud/admin/k6/loadtestrun/view/7416244/?_changelist_filters=q%3D2736043

---

> [!NOTE]
> **High Risk**
> Adds secrets configuration (including a run-scoped token) to `TestRun` status/CRD and propagates it into runner pod environment, which is security-sensitive and could expose credentials via Kubernetes status storage/logging. Also changes cloud create-test-run response parsing, so mismatches with the cloud API contract could break cloud-output runs.
> 
> **Overview**
> Cloud-output test runs now capture a secrets configuration returned by the cloud `POST /v1/tests` response and persist it on the `TestRun` status as **new field** `status.secretsVars` (CRD/schema updated accordingly).
> 
> The controller encodes `secrets_config` + `test_run_token` into this status field during initialization, and runner job creation decodes it to inject the corresponding secret-source env vars. `CreateTestRun` was refactored to parse these additional response fields (with tests for backwards compatibility when secrets fields are absent) and new helpers/tests were added for encoding/decoding the secrets config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 240a2ad949f1a7e756ad1bdad8d0d5830ed6f38c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/api/v1alpha1/testrun_types.go
+++ b/api/v1alpha1/testrun_types.go
@@ -180,6 +180,7 @@ type TestRunStatus struct {
 	Stage           Stage  `json:"stage,omitempty"`
 	TestRunID       string `json:"testRunId,omitempty"`
 	AggregationVars string `json:"aggregationVars,omitempty"`
+	SecretsVars     string `json:"secretsVars,omitempty"`
 
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
--- a/internal/controller/k6_initialize.go
+++ b/internal/controller/k6_initialize.go
@@ -180,12 +180,7 @@ func SetupCloudTest(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, 
 			v1alpha1.UpdateCondition(k6, v1alpha1.CloudTestRunCreated, metav1.ConditionTrue)
 
 			k6.GetStatus().AggregationVars = cloud.EncodeAggregationConfig(testRunData.ConfigOverride)
-
-			if trData, err := cloud.GetTestRunData(r.k6CloudClient, testRunData.ReferenceID); err != nil {
-				log.Error(err, "Failed to fetch test run data for secrets config")
-			} else {
-				k6.GetStatus().SecretsVars = cloud.EncodeSecretsConfig(trData.SecretsConfig, trData.SecretsToken)
-			}
+			k6.GetStatus().SecretsVars = cloud.EncodeSecretsConfig(testRunData.SecretsConfig, testRunData.SecretsToken)
 
 			_, err := r.UpdateStatus(ctx, k6, log)
 			if err != nil {
--- a/internal/controller/k6_initialize.go
+++ b/internal/controller/k6_initialize.go
@@ -181,6 +181,12 @@ func SetupCloudTest(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, 
 
 			k6.GetStatus().AggregationVars = cloud.EncodeAggregationConfig(testRunData.ConfigOverride)
 
+			if trData, err := cloud.GetTestRunData(r.k6CloudClient, testRunData.ReferenceID); err != nil {
+				log.Error(err, "Failed to fetch test run data for secrets config")
+			} else {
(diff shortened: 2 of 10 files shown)


vortegatorres:
Test https://admin-prod-us-east-0.grafana.net/directory-k6-cloud/admin/k6/loadtestrun/view/7410899/?_changelist_filters=q%3D2736043 was executed using this branch.

pkg/cloud/cloud_output.go
```diff
@@ -82,24 +90,34 @@ func CreateTestRun(opts InspectOutput, instances int32, host, token string, log
 
 // We cannot use cloudapi.TestRun struct and cloudapi.Client.CreateTestRun call because they're not aware of
 // process_thresholds argument; so let's use custom struct and function instead
-func createTestRun(client *cloudapi.Client, host string, testRun *TestRun) (*cloudapi.CreateTestRunResponse, error) {
+func createTestRun(client *cloudapi.Client, host string, testRun *TestRun) (*CreateTestRunResult, error) {
```
yorugac, line 93: Why is it a new struct definition instead of importing the existing one? I'd much prefer to reuse everything possible. AFAIK, it should be possible.

pkg/cloud/secrets.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// EncodeSecretsConfig encodes the secrets configuration into a pipe-separated
+// string for storage in TestRun.Status.SecretsVars.
+// Returns an empty string when cfg is nil.
+func EncodeSecretsConfig(cfg *SecretsConfig, token string) string {
```
yorugac, line 12: Similar question to the struct one: does this function not exist anywhere else? Like in k6 CLI.

api/v1alpha1/testrun_types.go
```diff
@@ -180,6 +180,7 @@ type TestRunStatus struct {
 	Stage           Stage  `json:"stage,omitempty"`
 	TestRunID       string `json:"testRunId,omitempty"`
 	AggregationVars string `json:"aggregationVars,omitempty"`
+	SecretsVars     string `json:"secretsVars,omitempty"`
```
yorugac, line 183: 😔 Can't say I like this. Well, I don't like aggregation vars either, TBH.

yorugac (changes requested):
Hi @vortegatorres, 
I've done a preliminary sweep over the changes and would like to share this feedback. I feel like there's insufficient re-use of what was done in k6 CLI. Could you please double-check? I usually avoid duplicating as much as possible in k6-operator: the less code - the better.

internal/controller/k6_initialize.go
```diff
@@ -180,6 +180,7 @@ func SetupCloudTest(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun,
 			v1alpha1.UpdateCondition(k6, v1alpha1.CloudTestRunCreated, metav1.ConditionTrue)
 
 			k6.GetStatus().AggregationVars = cloud.EncodeAggregationConfig(testRunData.ConfigOverride)
+			k6.GetStatus().SecretsVars = cloud.EncodeSecretsConfig(testRunData.SecretsConfig, testRunData.SecretsToken)
```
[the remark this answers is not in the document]
yorugac: What can this token do again, @vortegatorres ? I.e. can we guarantee that it can operate only with this test run id and nothing else? It cannot be used to modify in any way another test run id and it cannot be used to create / delete another test run. 

Is this correct?

pkg/cloud/cloud_output.go
yorugac: This wasn't resolved 😅

PR #810, opened 2026-06-03
docs: Create a guide for migration from one instance setup to multiple inst...

### LINKED ISSUE #96: Create a guide for migration from one instance setup to multiple instance setup
Users of k6-operator can face migration issues where they need to re-configure both their setups and k6 scripts to distributed execution. Currently it's fully user's responsibility but a straight-forward guide on how to wrangle [k6 executors](https://k6.io/docs/using-k6/scenarios/executors/) into obedience without full understanding of their logic can be beneficial too. A related issue was raised in https://github.com/grafana/k6-operator/issues/90.

In general, this can be solved by re-computing number of VUs, e.g. if one was running `X` independent instances with `Y` VUs, for multiple instances setup they'll need `X * Y` VUs and parallelism set to `X`. This may be seen as basic arithmetic; however, a specific guide can both help people to operate between 1 instance VS N instances more easily and also serve as additional introduction into executors and execution segments.

arnavnagzirkar:
## Summary

The issue requested a documentation guide to help users migrate from a single k6 instance setup to a distributed multi-instance setup using k6-operator. No such guide existed.

## Root cause

The issue requested a documentation guide to help users migrate from a single k6 instance setup to a distributed multi-instance setup using k6-operator. No such guide existed. Users had to figure out VU/rate scaling and executor behavior on their own.

## What changed

Added `docs/migration-guide.md` with a step-by-step migration guide covering:

- How k6-operator distributes load using execution segments
- The core formula: total VUs = VUs per instance * parallelism
- Before/after examples for common executor types (vus, stages, constant-arrival-rate)
- What does not change (thresholds, setup/teardown behavior)
- A summary table for quick reference

## Issue

Fixes #96

Issue: https://github.com/grafana/k6-operator/issues/96

## Diffstat

```
docs/migration-guide.md | 133 ++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 133 insertions(+)
```

## Testing

- Ran the relevant tests and linter for the changed files while developing.

- Kept the change minimal and focused on this one issue.

## AI assistance

I used GitHub Copilot to help write parts of this change. I've reviewed and tested it myself, I understand what it does, and I'll follow up on any review feedback.

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -10,11 +10,11 @@ This means all runners execute the **same script** but each covers a different s
 
 ## Single instance setup
 
-A typical single-instance run:
+A typical single-instance run drives all virtual users from one pod:
 
 ```js
 export const options = {
-  vus: 50,
+  vus: 200,
   duration: '1m',
 };
 ```
@@ -34,19 +34,19 @@ spec:
       file: test.js
 ```
 
-This starts 50 VUs from a single pod running for 1 minute.
+This starts 200 VUs from a single pod running for 1 minute.
 
 ## Migrating to multiple instances
 
-To spread the same load across multiple runners, increase `parallelism` and scale up VUs proportionally. The formula is simple:
+The usual reason to move to multiple runners is to spread an existing load across more pods, for example when a single pod no longer has enough CPU or memory to drive all the VUs. With k6-operator you keep the same values in your script options and only increase `parallelism`. The operator divides the total load across the runners for you using execution segments, so you do not multiply anything yourself.
 
-> **total VUs = original VUs per instance * number of instances**
+> **The values in your script options are always the total load. `parallelism` only controls how many runner pods share that total.**
 
-So if you were running 1 instance with 50 VUs and you want to scale to 4 instances:
+So to move the 200 VU test above onto 4 runners, leave the script unchanged and set `parallelism: 4`:
 
 ```js
 export const options = {
-  vus: 200,   // 50 * 4
+  vus: 200,   // unchanged: this is the total across all runners
   duration: '1m',
 };
 ```
@@ -64,11 +64,15 @@ spec:
       file: test.js
 ```
 
-Each runner handles 50 VUs, for a total of 200 VUs across the cluster. This replicates the original load profile on each node.
+Each runner pod now handles about 50 VUs (200 / 4), for the same total of 200 VUs across the cluster. The load profile is unchanged, it is just distributed across pods.
+
+### Increasing total load at the same time
+
+If you also want to drive more load than before, raise the values in your script options. For example, setting `vus: 400` with `parallelism: 4` runs about 100 VUs per runner for 400 VUs total. Changing the total load and changing `parallelism` are independent decisions: `parallelism` distributes whatever total you configure.
 
 ## Working with other executors
 
-The same principle applies to other [k6 executors](https://grafana.com/docs/k6/latest/using-k6/scenarios/executors/).
+The same rule applies to other [k6 executors](https://grafana.com/docs/k6/latest/using-k6/scenarios/executors/): the values you set are the total, and k6-operator splits them across runners.
 
 ### constant-arrival-rate
 
@@ -77,7 +81,7 @@ export const options = {
   scenarios: {
     contacts: {
       executor: 'constant-arrival-rate',
-      rate: 400,          // 100 * 4 runners
+      rate: 400,          // total request rate across all runners
       timeUnit: '1s',
       duration: '1m',
       preAllocatedVUs: 80,
@@ -87,20 +91,20 @@ export const options = {
 };
 ```
 
-With `parallelism: 4`, each runner fires 100 requests per second, for 400 req/s total.
+With `parallelism: 4`, each runner fires about 100 requests per second, for 400 req/s total.
 
 ### ramping-vus
 
 ```js
 export const options = {
   stages: [
-    { target: 400, duration: '30s' },  // ramps up to 400 total (100 per runner)
+    { target: 400, duration: '30s' },  // total target across all runners
     { target: 0,   duration: '30s' },
   ],
 };
 ```
 
-k6 splits the VU ramp proportionally between runners using execution segments, so the total ramp shape is preserved.
+k6 splits the VU ramp proportionally between runners using execution segments, so each runner ramps toward about 100 VUs while the total ramp shape is preserved.
 
 ## What stays the same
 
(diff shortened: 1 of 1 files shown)


yorugac:
Hi @arnavnagzirkar, thanks for the PR. I'll be able to look at it in detail next week. But [this issue](https://github.com/grafana/k6-operator/issues/96) is from the times when we kept all the docs here, in this repo. Which is no longer the case. So we might want to move this to k6-docs instead. cc @heitortsergent WDYT? Should we put this guide there, in k6-operator section?

arnavnagzirkar:
Hi @yorugac, thanks for taking a look and for explaining the docs move.

That makes sense to me. Putting this in k6-docs under the distributed k6 / k6-operator section sounds like the right home. From a quick look, `docs/sources/k6/next/set-up/set-up-distributed-k6/usage/` seems like a natural fit, next to pages like `executing-k6-scripts-with-testrun-crd.md` and `scheduling-tests.md`.

I'm happy to port the guide there as a new page in that section, or fold it into an existing page if you'd prefer. Once you and @heitortsergent settle on the exact destination, I'll open a PR in k6-docs and close this one in favor of it.

No rush, I'll wait for your steer next week. Thanks again.

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
+# Migrating from single instance to multiple instances
+
+This guide describes how to migrate an existing k6 setup from a single runner to multiple runners using k6-operator. It also explains the relationship between k6 executors and execution segments, which is important for understanding how distributed runs work.
```
yorugac, line 3:
```suggestion
This guide describes how to migrate an existing k6 setup from a single runner to multiple runners using k6-operator.
```
Let's stay as focused as possible.

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+
+## How k6-operator distributes a test
+
+When you set `parallelism: N` in a `TestRun`, k6-operator starts `N` runner pods. Each runner receives an [execution segment](https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/#execution-segment) that limits it to a portion of the total work. For example, with `parallelism: 2`, runner 0 gets segment `0:1/2` and runner 1 gets segment `1/2:1`.
+
+This means all runners execute the **same script** but each covers a different slice of the virtual users and iterations. The total load across all runners equals what you define in the script options.
```
yorugac, line 9:
```suggestion
This means all runners execute the **same script** but each runner gets a chunk of VUs and iterations to execute. The total load across all runners equals what you define in the script options.
```

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Single instance setup
+
+A typical single-instance run drives all virtual users from one pod:
```
yorugac, line 13:
```suggestion
A typical single-instance run drives all VUs from one pod:
```
I'm also not sure about this "drives"... @heitortsergent is best to make stylistic corrections though.

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
+# Migrating from single instance to multiple instances
+
+This guide describes how to migrate an existing k6 setup from a single runner to multiple runners using k6-operator. It also explains the relationship between k6 executors and execution segments, which is important for understanding how distributed runs work.
+
+## How k6-operator distributes a test
```
yorugac, line 5: This section should also refer to the existing docs for k6-operator:
https://grafana.com/docs/k6/latest/set-up/set-up-distributed-k6/usage/configure-testrun-crd/#parallelism

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+  parallelism: 4
+  script:
+    configMap:
+      name: k6-test
+      file: test.js
+```
+
+Each runner pod now handles about 50 VUs (200 / 4), for the same total of 200 VUs across the cluster. The load profile is unchanged, it is just distributed across pods.
+
+### Increasing total load at the same time
```
yorugac, line 69: This section seems a bit redundant to me. I'd remove it.

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+
+If you also want to drive more load than before, raise the values in your script options. For example, setting `vus: 400` with `parallelism: 4` runs about 100 VUs per runner for 400 VUs total. Changing the total load and changing `parallelism` are independent decisions: `parallelism` distributes whatever total you configure.
+
+## Working with other executors
+
+The same rule applies to other [k6 executors](https://grafana.com/docs/k6/latest/using-k6/scenarios/executors/): the values you set are the total, and k6-operator splits them across runners.
```
yorugac, line 75:
```suggestion
The same rule applies to other [k6 executors](https://grafana.com/docs/k6/latest/using-k6/scenarios/executors/).
```

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+  stages: [
+    { target: 400, duration: '30s' },  // total target across all runners
+    { target: 0,   duration: '30s' },
+  ],
+};
+```
+
+k6 splits the VU ramp proportionally between runners using execution segments, so each runner ramps toward about 100 VUs while the total ramp shape is preserved.
+
+## What stays the same
```
yorugac, line 109: I'm not sure what source of info you use, but both points in this section are not correct... So it's either fix them or remove the section 😅 

(just in case, it's fine to remove it for now: we can expand on it at a later point too.)

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Summary
+
+To distribute an existing test across more runners, keep your script options the same and increase `parallelism`:
+
+| Goal | Original setup | Multi-instance setup |
+|---|---|---|
+| Same total load, distributed | `parallelism: 1`, `vus: Y` | `parallelism: X`, `vus: Y` (unchanged) |
+| Same total request rate, distributed | `parallelism: 1`, `rate: R` | `parallelism: X`, `rate: R` (unchanged) |
+| Same total ramp, distributed | `parallelism: 1`, stages ramp to `T` | `parallelism: X`, stages ramp to `T` (unchanged) |
```
yorugac, line 137: This table may be misleading, because it references the examples above but without being explicit. Maybe remove it? Does it add new info to what is written above?

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+
+The key rule: the VU or rate values in your script are always the total. To distribute an existing test across more runner pods, keep those values the same and increase `parallelism`. Only change the values when you actually want to change the total load. k6-operator takes care of the split by assigning each runner an execution segment.
```
yorugac, line 139:
```suggestion
The key rule: the VU or rate values in your script are always the total. To distribute an existing test across more runner pods, keep those values the same and increase `parallelism`. Only change the values in the script when you actually want to change the total load. k6-operator takes care of the split by assigning each runner an execution segment.
```

docs/migration-guide.md
```diff
@@ -0,0 +1,139 @@
...
+      timeUnit: '1s',
+      duration: '1m',
+      preAllocatedVUs: 80,
+      maxVUs: 400,
+    },
+  },
+};
+```
+
+With `parallelism: 4`, each runner fires about 100 requests per second, for 400 req/s total.
```
yorugac, line 94: IIRC, this may not be fully correct. I suggest to dig into this in great detail so that we avoid false claims.

yorugac (changes requested):
Hi @arnavnagzirkar,
I've done a preliminary pass over this doc, highlighting the most noticeable points to change. We've also chatted with @heitortsergent about this a bit and agreed that this guide should be in k6-docs. 
@arnavnagzirkar, could you please address my comments and please move this PR there? We can continue in k6-docs then.

docs/migration-guide.md
yorugac: Actually, it might be easiest to ask for clarification / review from k6-core on this part. Once PR is moved to k6-docs.

PR #820, opened 2026-06-18
Hardening memory and concurrency

yorugac:
A few hardening fixes here:
- fixing missing / drifted away HTTP `body.Close()` calls;
- hardening the PLZ poller against misbehaving handler calls: that solves a potential data race; 
    - e.g. issues on GCk6 or k8s can delay healthy processing of the test run: that shouldn't hang the poller or finalization
- introducing centralized context in the PLZ poller and using that to cancel all the calls in the handler properly.

---

> [!NOTE]
> **Medium Risk**
> Changes PLZ shutdown and in-flight TestRun creation behavior during deregistration; incorrect cancellation could skip or abort k8s work, though scope is limited to the poller/PLZ path.
> 
> **Overview**
> **HTTP cleanup:** k6 runner readiness checks now **`defer resp.Body.Close()`** immediately after a successful `http.Get` in **`isServiceReady`** and **`isJobRunning`**, including moving the close in the stopped-jobs path so the body is always released even when status or JSON parsing fails.
> 
> **PLZ poller concurrency:** **`conn.Poller`** no longer stops via a blocking **`done` channel send**. It uses a **mutex**, **pointer receivers**, a **`context.Context`** created on **`Start`**, and **`Stop`** that **cancels the context** and returns immediately (idempotent, safe if **`OnInterval`** is stuck). Callers can use **`Context()`** to abort downstream work when the poller stops.
> 
> **Downstream wiring:** The cloud **test run poller** sends IDs to its channel with a **select** against **`ctx.Done()`** so shutdown does not block on a full or slow consumer. **`PLZWorker.handle`** takes that context for **`Get`/`Create`** on TestRun CRs instead of **`context.Background`**, and tests cover normal stop, stop while the interval handler deadlocks, and context cancellation on stop.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48ddd3d5915f60e92a3088381f7159ae5652e72e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/controller/k6_start.go
+++ b/internal/controller/k6_start.go
@@ -23,6 +23,7 @@ func isServiceReady(log logr.Logger, service *v1.Service) bool {
 		log.Error(err, fmt.Sprintf("failed to get status from %v", service.Name))
 		return false
 	}
+	defer resp.Body.Close() //nolint:errcheck
 
 	return resp.StatusCode < 400
 }
--- a/internal/controller/k6_stopped_jobs.go
+++ b/internal/controller/k6_stopped_jobs.go
@@ -22,6 +22,7 @@ func isJobRunning(log logr.Logger, service *v1.Service) bool {
 	if err != nil {
 		return false
 	}
+	defer resp.Body.Close() //nolint:errcheck
 
 	// Response has been received so assume the job is running.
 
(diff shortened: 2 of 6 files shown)


yorugac:
cursor review

dgzlopes (approved):

PR #828, opened 2026-06-26
fix(helm): trim generated resource names to 63 chars

### LINKED ISSUE #660: Trim service names to be under 63 chars
### Brief summary

The `controller-manager service` defined [here](https://github.com/grafana/k6-operator/blob/main/charts/k6-operator/templates/service.yaml) has a really long name which along with the `fullname` goes beyond 63 chars and fails with `Invalid value: "...": must be no more than 63 characters`.

Although there is a way to proceed by specifying a really small `fullnameOverride`, the chart should handle such issues by trimming the name automatically.

1.0.0

1.0.0

n/a

_No response_

Install k6 operator using helm with fullnameOverride set to `synthetics-dev-usw2-01` (22 chars).

### Expected behaviour

k6 operator should be installed successfully!

### Actual behaviour

fails with `Invalid value: "synthetics-dev-usw2-01-k6-operator-controller-manager-metrics-service": must be no more than 63 characters`.

(final service name has 69 chars)

moko-poi:
Fixes #660

## Problem

The `k6-operator.fullname` helper is already truncated to 63 chars, but the templates append a suffix to it (e.g. `-controller-manager-metrics-service`, 35 chars). When the release/fullname is long enough, the final resource name exceeds the Kubernetes 63-char name limit and `helm install` fails:

```
Invalid value: "synthetics-dev-usw2-01-k6-operator-controller-manager-metrics-service": must be no more than 63 characters
```

Today the only workaround is setting a very short `fullnameOverride`. As noted in the issue, the chart should handle this automatically.

## Change

- Rework `k6-operator.fullname` to take an optional `suffix` via a dict argument: `include "k6-operator.fullname" (dict "context" $ "suffix" "-manager-role")`.
- The fullname part (not the suffix) is truncated to `63 - len(suffix)`, so the whole name always fits in 63 chars while the suffix is always preserved. Keeping the suffix is what distinguishes the otherwise identical Role/RoleBinding names, so truncating it away would collapse several RBAC objects onto one name.
- Use the helper for every resource name that builds on the fullname: service, deployment, the roles/cluster-roles and their (cluster-)role-bindings, and the leader-election role/binding. Role / RoleBinding pairs use the same helper + the same suffix, so their `roleRef` names stay consistent with the role definitions even when truncation kicks in.
- Per review feedback, this is a single helper (no separate `fullnameWithSuffix`); the non-suffix call sites (`serviceAccountName`, namespace label) use the same dict form without a suffix.

## Verification

Rendered the chart with the reproducing values from the issue (`fullnameOverride=synthetics-dev-usw2-012345678`, `service.enabled=true`):

- Previously-overflowing service name (65 chars) is now exactly 63 chars with the suffix intact; **all** generated names are ≤ 63.
- Every `roleRef` name still matches a defined role name (verified under a stress release name long enough to force role-name truncation too).
- `helm template` with the default release name produces an **identical** manifest to `main` (no regression for existing installs).
- `helm lint` passes.

No chart version bump (left to maintainers), and `values.yaml` is unchanged so no docs/schema regeneration is needed.

## Note for release notes

When the release name (or `fullnameOverride`) is long enough that truncation applies, the rendered resource names change on upgrade (e.g. the Deployment is recreated under the new, truncated name). Default release names are unaffected. Users who reference these resource names in monitoring/alerting should update them.

### DIFF
--- a/charts/k6-operator/templates/_helpers.tpl
+++ b/charts/k6-operator/templates/_helpers.tpl
@@ -7,37 +7,31 @@ Expand the name of the chart.
 
 
 {{/*
-Create a default fully qualified app name.
-We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+Create a default fully qualified app name, optionally appending a suffix.
+The whole name is kept within the Kubernetes 63-char name limit (DNS naming
+spec): the fullname part (not the suffix) is truncated, so the suffix is
+always preserved. This keeps names that share a fullname but differ only by
+suffix (e.g. the various Roles/RoleBindings) unique.
 If release name contains chart name it will be used as a full name.
+Usage: {{ include "k6-operator.fullname" (dict "context" $ "suffix" "-controller-manager") }}
+       {{ include "k6-operator.fullname" (dict "context" $) }}
 */}}
 {{- define "k6-operator.fullname" -}}
-{{- if .Values.fullnameOverride }}
-{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- $name := default .Chart.Name .Values.nameOverride }}
-{{- if contains $name .Release.Name }}
-{{- .Release.Name | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
-{{- end }}
-{{- end }}
-{{- end }}
-
-{{/*
-Create a resource name by appending a suffix to the fullname, keeping the whole
-name within the Kubernetes 63-char name limit (DNS naming spec).
-The fullname is already truncated to 63 chars, but appending a suffix can push
-the resulting name over the limit. The fullname part (not the suffix) is
-truncated so the suffix is always preserved; this keeps names that share a
-fullname but differ only by suffix (e.g. the various Roles/RoleBindings) unique.
-Usage: {{ include "k6-operator.fullnameWithSuffix" (dict "context" $ "suffix" "-controller-manager") }}
-*/}}
-{{- define "k6-operator.fullnameWithSuffix" -}}
-{{- $suffix := .suffix -}}
-{{- $base := include "k6-operator.fullname" .context | trunc (int (sub 63 (len $suffix))) | trimSuffix "-" -}}
-{{- printf "%s%s" $base $suffix }}
-{{- end }}
+{{- $ctx := .context -}}
+{{- $suffix := .suffix | default "" -}}
+{{- $base := "" -}}
+{{- if $ctx.Values.fullnameOverride -}}
+{{- $base = $ctx.Values.fullnameOverride -}}
+{{- else -}}
+{{- $name := default $ctx.Chart.Name $ctx.Values.nameOverride -}}
+{{- if contains $name $ctx.Release.Name -}}
+{{- $base = $ctx.Release.Name -}}
+{{- else -}}
+{{- $base = printf "%s-%s" $ctx.Release.Name $name -}}
+{{- end -}}
+{{- end -}}
+{{- printf "%s%s" ($base | trunc (int (sub 63 (len $suffix))) | trimSuffix "-") $suffix -}}
+{{- end -}}
 
 {{/*
 Create chart name and version as used by the chart label.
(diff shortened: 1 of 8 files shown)


charts/k6-operator/templates/_helpers.tpl
```diff
@@ -24,6 +24,21 @@ If release name contains chart name it will be used as a full name.
...
+name within the Kubernetes 63-char name limit (DNS naming spec).
+The fullname is already truncated to 63 chars, but appending a suffix can push
+the resulting name over the limit. The fullname part (not the suffix) is
+truncated so the suffix is always preserved; this keeps names that share a
+fullname but differ only by suffix (e.g. the various Roles/RoleBindings) unique.
+Usage: {{ include "k6-operator.fullnameWithSuffix" (dict "context" $ "suffix" "-controller-manager") }}
+*/}}
+{{- define "k6-operator.fullnameWithSuffix" -}}
+{{- $suffix := .suffix -}}
+{{- $base := include "k6-operator.fullname" .context | trunc (int (sub 63 (len $suffix))) | trimSuffix "-" -}}
```
yorugac: It seems like `k6-operator.fullname` is now being called only in 2 places (other than here): service account and namespace. Would it be possible to rewrite this further to have just 1 function instead of 2?

yorugac (changes requested):
Thank you, @moko-poi for working on this!

The change seems OK overall to me, but I wonder if it can be simplified a bit more: please see my comment.

----
A note on risks during upgrades (after merging this):

> [cursor bugbot] Renames rendered resources when truncation applies, which can affect upgrades from long release names; default release names are unchanged per PR verification.

Preliminary. In practice this could mean a rename of Deployment for a subset of users: ones who already had a long release name. So specifically users from https://github.com/grafana/k6-operator/issues/660 who deploy with long name and `service.enabled: false`, assuming they do that. If the cluster is healthy, it should be pretty fast anyway. Just a bit worse than usual rolling update.

One more potential risk would be if someone also relies on that name in their o11y... Not sure how likely it is. But just in case, it might make sense to highlight this bit in the release notes.

charts/k6-operator/templates/_helpers.tpl
moko-poi: Good point, thank you! Merged the two helpers into a single `k6-operator.fullname` that takes an optional `suffix` via a dict argument (b2bd78a). The two non-suffix call sites (`serviceAccountName` and the namespace label) now use the same dict form. Rendered output is unchanged — verified `helm template` output is byte-identical before/after for the default release name, a 53-char release name, and a long `fullnameOverride`.

charts/k6-operator/templates/_helpers.tpl
```diff
@@ -24,6 +24,17 @@ If release name contains chart name it will be used as a full name.
...
+{{/*
+Create a resource name by appending a suffix to the fullname, truncated to 63
+chars to comply with the Kubernetes name length limit (DNS naming spec).
+The fullname is already truncated to 63 chars, but appending a suffix can push
+the resulting name over the limit, so the whole name is truncated again here.
+Usage: {{ include "k6-operator.fullnameWithSuffix" (dict "context" $ "suffix" "-controller-manager") }}
+*/}}
+{{- define "k6-operator.fullnameWithSuffix" -}}
+{{- printf "%s%s" (include "k6-operator.fullname" .context) .suffix | trunc 63 | trimSuffix "-" }}
+{{- end }}
```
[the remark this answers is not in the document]
moko-poi: This was true of the earlier commit (793e968), which truncated the combined string. Since 883ce0d the fullname part is truncated to `63 - len(suffix)` *before* the suffix is appended, so the suffix is always preserved and the Role/RoleBinding names remain distinct.

yorugac (approved):
Thank you for the updates, @moko-poi! 🙌 🚀

yorugac:
Oh, I didn't notice at first, but signatures are missing:
image
@moko-poi, could you please sign the commits? This is a requirement across Grafana org nowadays.

moko-poi:
@yorugac I re-signed the two commits, so they now show as verified. No content changes were made, just the signatures. Thanks again for the review!

yorugac:
Thanks, @moko-poi!

For the record, `govunlcheck` failure is due to newest Go CVE: appeared in parallel, not related to Helm changes here.

PR #841, opened 2026-07-02
fix(controller): return error when stop job creation fails

### LINKED ISSUE #840: bug: failed stop job creation leaves TestRun permanently stuck
## Bug

When `StopJobs` fails to create the stop job (e.g. due to a transient API server error, quota issue, or RBAC problem), the error is logged but the function returns `ctrl.Result{}, nil` — telling controller-runtime "success, do not requeue".

Since the `TestRun` stage was never updated to `stopped`, nothing else triggers another reconcile, and the object is permanently stuck with no user-visible indication of why.

## Affected file

`internal/controller/k6_stop.go`, lines 53-56:

```go
if err = r.Create(ctx, stopJob); err != nil {
    log.Error(err, "Failed to launch k6 test stop job.")
    return res, nil   // res is zero value — no requeue, nil error
}
```

`res` is never assigned in this function, so it is always the zero `ctrl.Result{}`. Returning `nil` as the error signals success to controller-runtime. The reconciler is not retried, and the `TestRun` cannot progress.

## Expected behaviour

A transient failure creating the stop job should be retried. The error should be returned so controller-runtime re-enqueues the reconcile request.

## Fix

- Return `err` instead of `nil` on unexpected Create errors so the reconciler retries.
- Handle `IsAlreadyExists` gracefully (covers the case where the reconciler fires twice before the status update completes), removing the existing `// TODO: add a check for existence of stop job` comment.

```go
if err = r.Create(ctx, stopJob); err != nil {
    if !k8sErrors.IsAlreadyExists(err) {
        log.Error(err, "Failed to launch k6 test stop job.")
        return res, err
    }
    log.Info("Stop job already exists, continuing")
}
```

jainik-s-patel:
## What

Fixes a bug in `StopJobs` where a failed stop job creation silently abandoned the reconcile loop, leaving the `TestRun` permanently stuck.

Fixes #840

## Why

`StopJobs` was returning `res, nil` on `Create` failure, where `res` is the zero `ctrl.Result{}`. This signals \"success, do not requeue\" to controller-runtime. Since the `TestRun` stage never advanced to `stopped`, no further reconcile was triggered and the object was stuck indefinitely with no user-visible error.

## Changes

`internal/controller/k6_stop.go`:
- Return `err` instead of `nil` on unexpected `Create` errors so the reconciler retries
- Handle `IsAlreadyExists` gracefully — covers the case where the reconciler fires twice before the status update completes
- Remove the `// TODO: add a check for existence of stop job` comment

## Checklist

- [x] `make fmt` — clean
- [x] `make vet` — clean
- [x] `make test` — all tests pass
- [x] `make lint` — 0 issues


yorugac:
The issue #840 is pending clarification, and this PR was opened immediately after the issue. Please avoid opening large and structure-changing PRs without a confirmed use case. I'm closing it for now.

jainik-s-patel:
You're right, I apologize. I found this through code reading without a confirmed reproduction. I should have asked for feedback on the issue before opening the PR. I'll make sure to wait for issue confirmation before submitting code changes going forward.

Issue #616, opened 2025-07-29
Refactor `api/` package


yorugac:
Currently some definitions in `api/` package contain dependency on k6 package, like `"go.k6.io/k6/cloudapi"`. This can cause trouble in some code vendoring setups, when importing the package. Refactor the `api/` package so that it doesn't have such dependencies.

PR #849, opened 2026-07-14
Pass ephemeral test run token as `K6_CLOUD_TOKEN` for PLZ tests

### LINKED ISSUE #823: Introduce two-level system of authentication into `PrivateLoadZone`
### Feature Description

Currently `PrivateLoadZone` is working with 1-level of authentication: the stack token that is configured during PLZ creation ([docs](https://grafana.com/docs/grafana-cloud/testing/k6/author-run/private-load-zone/)) is used to execute all the PLZ tests.

That brings up certain security concerns when users of PLZ must have different levels of access to the cluster and to GCk6. We've previously hidden the stack token from plain sight with [`valueFrom`](https://github.com/grafana/k6-operator/issues/615) (v0.0.23+), so that users who can view Pods but not Secrets cannot access the stack token that way.

Going further, it makes sense to introduce a 2-level system of authentication:
- PLZ stays configured with a stack token;
- PLZ tests are executed with another, short-lived token having smaller scope of permissions than a stack token.

The new token will likely be passed from GCk6 as `gck6_env_vars.K6_CLOUD_TOKEN` as described in https://github.com/grafana/k6-operator/issues/742. Passing env var to k6 is trivial, but it will have to be passed as a string which _makes its value more easily readable again_, effectively reverting the change in https://github.com/grafana/k6-operator/issues/615.

However, GCk6 is going to guarantee that the new token is very limited in scope, can be used only with one given test run ID and will have ephemeral lifetime. To be precise, there shouldn't be any action unrelated to the given test run that's possible to do with the new ephemeral token.

https://github.com/grafana/k6-operator/issues/742 must be implemented first.

I'm splitting this into a separate issue from https://github.com/grafana/k6-operator/issues/742, as this change must also come with the updates of logic around tokens, to ensure that different `TestRun`s all operate on different tokens too (in PLZ mode).

yorugac:
Fixes: https://github.com/grafana/k6-operator/issues/823

A small refactoring included on the way, to improve readability and as a prep for #616 (creating an internal TestRun type) and #668 (support for new k6 command and new auth), plus:

- `cloud.TokenInfo` -> `cloud.SecretTokenInfo`, because we now can have more than one source of the token (k8s Secret vs GCk6)
- legacy var renaming: `k6` -> `tr`

Note: cloud output mode remains as it was before. The remake of cloud output logic will come in / after the https://github.com/grafana/k6-operator/issues/668. However, there is an expectation that a user of cloud output is not setting `K6_CLOUD_TOKEN` env var to runners: that probably won't work. (This wasn't supported prior to this PR either, it just becomes more explicit now.)

### DIFF
--- a/pkg/resources/jobs/runner_test.go
+++ b/pkg/resources/jobs/runner_test.go
@@ -165,7 +165,7 @@ func Test_NewRunnerJob(t *testing.T) {
 	tests := []struct {
 		name             string
 		script           *types.Script
-		tokenInfo        *cloud.TokenInfo
+		sti              *cloud.SecretTokenInfo
 		setupTestRun     func(*v1alpha1.TestRun)
 		setupExpectedJob func(*batchv1.Job)
 	}{
@@ -250,8 +250,8 @@ func Test_NewRunnerJob(t *testing.T) {
 			},
 		},
 		{
-			name:      "cloud output mode",
-			tokenInfo: cloud.NewTokenInfo("", "").InjectValue("token"),
+			name: "cloud output mode",
+			sti:  cloud.NewSecretTokenInfo("", "").InjectValue("token"),
 			setupTestRun: func(k6 *v1alpha1.TestRun) {
 				k6.Spec.Arguments = "--out cloud"
 				k6.Spec.Runner.Metadata.Labels = nil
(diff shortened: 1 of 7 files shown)


pkg/resources/jobs/runner_test.go
```diff
@@ -526,6 +527,38 @@ func Test_NewRunnerJob(t *testing.T) {
 							},
 						},
 					},
+					{Name: "K6_CLOUD_PUSH_REF_ID", Value: "plz-run-123"},
```
yorugac: TODO: fix order

dgzlopes (approved):

PR #854, opened 2026-07-25
refactor: collect default values at the top of job builders

### LINKED ISSUE #784: Improvement Code Structuring | Default value of variables
**If you look in here**:
The default value of variables are mentioned neatly on top of the function
https://github.com/grafana/k6-operator/blob/0ea3316246a4575ba0e4187deeb32c53b5cb775b/pkg/resources/jobs/initializer.go#L20

**In other files**:
They are scattered across length of function.
https://github.com/grafana/k6-operator/blob/0ea3316246a4575ba0e4187deeb32c53b5cb775b/pkg/resources/jobs/runner.go#L87

**Please make changes in these files**:
https://github.com/grafana/k6-operator/blob/main/pkg/resources/jobs/runner.go
https://github.com/grafana/k6-operator/blob/main/pkg/resources/jobs/starter.go
https://github.com/grafana/k6-operator/blob/main/pkg/resources/jobs/stopper.go

moko-poi:
Closes #784

## What

`NewInitializerJob` declares all of its default values in a single `var` block at the top of the function:

https://github.com/grafana/k6-operator/blob/ffca562/pkg/resources/jobs/initializer.go#L22-L30

`NewRunnerJob` and `NewStarterJob` follow the same "default -> override -> pass to `PodSpec`" shape, but scatter their defaults across the function body: `runner.go` had 2 of 8 in the `var` block, `starter.go` had 1 of 7. This PR moves the scattered ones into the existing `var` blocks, in the same order as `initializer.go`, so all job builders read the same way.

`stopper.go` needs no change: `NewStopJob` builds on top of `NewStarterJob`, so it inherits the cleanup.

## Notes for review

- **Behaviour is unchanged.** Only declaration sites move; the override branches stay where they were.
- **The unit tests are untouched on purpose.** They compare the whole generated `Job` object, so any difference in the resulting `PodSpec` would fail them - which makes them the regression check for this change.
- Variables are not renamed (e.g. `starterAnnotations` is left as is rather than becoming `annotations`), to keep the diff limited to movement.
- `ports` in `runner.go` is now built with the single-line `append` form used in `initializer.go`, instead of a literal followed by a separate `append`.
- `var zero32 int32` is deliberately left alone: `initializer.go` and `starter.go` both declare it on its own just before the `Job` literal, so it is already consistent there.

Moving declarations earlier changes when they are evaluated, so I checked that nothing in between has side effects:
- `newLabels()` only returns a map literal.
- `k6.GetSpec().Ports` is read exactly once in `runner.go` and never written.
- `ports` starts from a fresh slice literal, so no backing array is shared with `k6.GetSpec().Ports`.

## Verification

- `go test ./pkg/resources/jobs/...` - ok, 45 tests pass
- `go test -race` over all packages except `e2e` - all pass
- `make lint` (golangci-lint with the kube-api-linter plugin, `./...`) - 0 issues
- `gofmt` / `go vet` / `go build ./...` - clean

### DIFF
--- a/pkg/resources/jobs/runner.go
+++ b/pkg/resources/jobs/runner.go
@@ -85,21 +85,24 @@ func NewRunnerJob(k6 *v1alpha1.TestRun, index int, tokenInfo *cloud.TokenInfo) (
 	command = script.UpdateCommand(command)
 
 	var (
-		zero32        int32 = 0
-		schedulerName       = corev1.DefaultSchedulerName
+		zero32                       int32 = 0
+		image                              = "grafana/k6:latest"
+		runnerAnnotations                  = make(map[string]string)
+		runnerLabels                       = newLabels(k6.NamespacedName().Name)
+		serviceAccountName                 = "default"
+		automountServiceAccountToken       = true
+		ports                              = append([]corev1.ContainerPort{{ContainerPort: 6565}}, k6.GetSpec().Ports...)
+		schedulerName                      = corev1.DefaultSchedulerName
 	)
 
-	image := "grafana/k6:latest"
 	if k6.GetSpec().Runner.Image != "" {
 		image = k6.GetSpec().Runner.Image
 	}
 
-	runnerAnnotations := make(map[string]string)
 	if k6.GetSpec().Runner.Metadata.Annotations != nil {
 		runnerAnnotations = k6.GetSpec().Runner.Metadata.Annotations
 	}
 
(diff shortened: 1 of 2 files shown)


yorugac (approved):
@moko-poi, thank you for tackling this!

A note. There's a failed govulncheck in CI, but it's a false positive in this PR: it was failing on all branches last week, until the security patch for libs was merged. IOW, it's not caused by this PR, and on main branch it should all be OK.

PR #856, opened 2026-07-27
fix: avoid recreating start and stop jobs

### LINKED ISSUE #310: Check existence of start and stop jobs before creation
### Feature Description

It's common to see the following images in k6-operator logs during test runs:
```
2023-10-13T19:32:16Z    ERROR   controllers.TestRun     Failed to launch k6 test starter        {"namespace": "default", "name": "k6-sample", "reconcileID": "a04a3862-895c-4b66-985d-f98d558c66c6", "error": "jobs.batch \"k6-sample-starter\" already exists"}
```
Both for start and stop job. This happens because there's no check for existence of those jobs prior to creating them and the state of TestRun hasn't caught up yet.

Unless something else comes up, it should a straight-forward addition.

_No response_

Guflly:
Fixes #310

## Problem

Reconciliation can reach start or stop job creation again before the TestRun status update has caught up. Creating the same job again then reports an `AlreadyExists` error.

## Change

Check for an existing job before creation and continue the status transition when it is already present. An `AlreadyExists` response from a concurrent creation is handled the same way.

## Verification

- `go test ./internal/controller -run '^TestCreateJobIfNotExists$' -count=10`
- `go test ./... -run '^TestCreateJobIfNotExists$' -count=1`
- `go vet ./internal/controller`

### DIFF
--- a/internal/controller/job_helper.go
+++ b/internal/controller/job_helper.go
@@ -1,0 +1,29 @@
+package controllers
+
+import (
+	"context"
+
+	batchv1 "k8s.io/api/batch/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func createJobIfNotExists(ctx context.Context, k8sClient client.Client, job *batchv1.Job) (bool, error) {
+	err := k8sClient.Get(ctx, client.ObjectKeyFromObject(job), &batchv1.Job{})
+	if err == nil {
+		return false, nil
+	}
+	if !apierrors.IsNotFound(err) {
+		return false, err
+	}
+
+	err = k8sClient.Create(ctx, job)
+	if apierrors.IsAlreadyExists(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
--- a/internal/controller/job_helper_test.go
+++ b/internal/controller/job_helper_test.go
@@ -1,0 +1,49 @@
+package controllers
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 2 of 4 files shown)


internal/controller/job_helper.go
```diff
@@ -0,0 +1,29 @@
...
+	err := k8sClient.Get(ctx, client.ObjectKeyFromObject(job), &batchv1.Job{})
+	if err == nil {
+		return false, nil
+	}
+	if !apierrors.IsNotFound(err) {
+		return false, err
+	}
+
+	err = k8sClient.Create(ctx, job)
+	if apierrors.IsAlreadyExists(err) {
```
yorugac, line 21: Can this case here be true in practice?

yorugac (changes requested):
Hi @Guflly, thanks for the PR!

Signing a CLA is a requirement: there's a message about it in the PR 🙂 

I have a question about logic - see the comments. And a couple of requests:
- Please rebase onto main: there was a security update there which should fix the failing govulncheck.
- Rename the file `k6_job` to `job_helper.go`. k6 prefix is an old convention which will likely be removed in the future.

Could you please check and clarify the above?

internal/controller/job_helper.go
Guflly: Yes. The controller-runtime client reads through the informer cache but writes directly to the API server. If the cache still returns NotFound after the Job has been created, Create sees the live object and returns AlreadyExists. MaxConcurrentReconciles is 1 here, but that cache propagation window can still happen.

Guflly:
Signed the CLA, rebased onto main, and renamed the helper files. I replied inline about the AlreadyExists case. Controller tests and go vet pass. The refreshed workflows are waiting for approval.

Guflly:
Rewrote the branch as one verified signed commit. The tree is unchanged.

yorugac (approved):
@Guflly, thank you for all the updates and clarifications 🙌

Issue #304, opened 2023-10-10
Support for restart of a runner


yorugac:
If runner pod is restarted before test execution is completed, the new runner does not receive a starting signal from starter and remains paused while the operator waits indefinitely.

It's unclear how often this happens in real-life scenarios but there's such a possibility. 

What the operator should do in such scenario? Some potential ideas:
1. ignore the restarted pod and try to finish the test with `n - 1` pods. Cons: it's not easy to implement correctly (without breaking existing functionality) if possible at all.
2. start the test on the restarted runner and let it finish. Cons: this will make the test longer than estimated and will skew the results in many cases.
3. fail the test on some timeout. This is the one that will likely get implemented as consequence of issue [#222](https://github.com/grafana/k6-operator/issues/222)

Another caveat is that there can be more than 1 restarting pod. In case of large failure of node group, there could be all runners failing, in theory. In which case, the 3rd option seems as the one making the most sense.

Opening this issue as a follow-up from [#138](https://github.com/grafana/k6-operator/issues/138). Feedback and thoughts are welcome.

PR #867, opened 2026-08-05
fix: handle OOMKilled runner pods

### LINKED ISSUE #251: Monitor runners in case of insufficient resources
### Feature Description

When one of the runners does not have sufficient resources allocated for the test, it goes into OOM state (insufficient memory for VUs. There can be other types of error for the same case as well). This condition is not monitored by the operator in any way, resulting in infinite wait loop for the pods to bootstrap.

This case should be monitored by the operator, followed by abortion of the test.

By initial experiments, there are two loops that can become infinite in such cases, at `stage = "created"` and `stage = "started"`.

Note that test runs in different modes need to handled this case differently.

Potentially connected issue: https://github.com/grafana/k6-operator/issues/222

Guflly:
Detects OOMKilled runner and init containers, including previous termination states.

OSS test runs move to the error stage. Cloud test runs send the OOM error and abort events. The check runs during startup and the setup, runner-finish, and teardown paths.

Tests: focused OOM tests, non-e2e unit tests, and non-e2e go vet.

Closes #251

### DIFF
--- a/internal/controller/common.go
+++ b/internal/controller/common.go
@@ -225,12 +225,18 @@ func handleRunnerOOMKilled(ctx context.Context, log logr.Logger, k6 *v1alpha1.Te
 	log.Info(msg)
 
 	if isCloudTestRun(k6) {
+		if v1alpha1.IsTrue(k6, v1alpha1.CloudTestRunAborted) {
+			return true, nil
+		}
 		if cloudClient == nil {
 			return true, errors.New("cloud client is not configured")
 		}
 		events := cloud.ErrorEvent(cloud.OOMError).WithDetail(msg).WithAbort()
-		cloud.SendTestRunEvents(cloudClient, k6.TestRunID(), log, events)
-		return true, nil
+		if err := cloud.SendTestRunEvents(cloudClient, k6.TestRunID(), log, events); err != nil {
+			return true, err
+		}
+		_, err := StopJobs(ctx, log, k6, r)
+		return true, err
 	}
 
 	v1alpha1.UpdateCondition(k6, v1alpha1.TestRunRunning, metav1.ConditionFalse)
--- a/internal/controller/common.go
+++ b/internal/controller/common.go
@@ -15,8 +15,10 @@ import (
 	"github.com/grafana/k6-operator/api/v1alpha1"
 	"github.com/grafana/k6-operator/pkg/cloud"
 	"github.com/grafana/k6-operator/pkg/testrun"
+	"go.k6.io/k6/v2/cloudapi"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -25,6 +27,7 @@ import (
 
 const (
 	errMessageTooLong = "Creation of %s takes too long: your configuration might be off. Check if %v were created successfully."
+	oomKilledReason   = "OOMKilled"
 )
 
 // It may take some time to retrieve inspect output so indicate with boolean if it's ready
@@ -191,4 +194,56 @@ func runTeardown(ctx context.Context, hostnames []string, log logr.Logger) {
 	if err := testrun.RunTeardown(ctx, hostnames); err != nil {
 		log.Error(err, "Failed to invoke teardown()")
 	}
+}
+
+func isOOMKilled(status corev1.ContainerStatus) bool {
+	return status.State.Terminated != nil && status.State.Terminated.Reason == oomKilledReason ||
+		status.LastTerminationState.Terminated != nil && status.LastTerminationState.Terminated.Reason == oomKilledReason
+}
+
+func findOOMKilledContainer(pods []corev1.Pod) (string, string, bool) {
+	for _, pod := range pods {
+		for _, statuses := range [][]corev1.ContainerStatus{pod.Status.InitContainerStatuses, pod.Status.ContainerStatuses} {
+			for _, status := range statuses {
+				if isOOMKilled(status) {
+					return pod.Name, status.Name, true
+				}
+			}
+		}
+	}
+
+	return "", "", false
+}
+
+func handleRunnerOOMKilled(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *TestRunReconciler, pods []corev1.Pod, cloudClient *cloudapi.Client) (bool, error) {
+	podName, containerName, found := findOOMKilledContainer(pods)
+	if !found {
+		return false, nil
+	}
+
+	msg := fmt.Sprintf("runner pod %s container %s was terminated with reason %s", podName, containerName, oomKilledReason)
+	log.Info(msg)
+
+	if isCloudTestRun(k6) {
+		if cloudClient == nil {
+			return true, errors.New("cloud client is not configured")
+		}
+		events := cloud.ErrorEvent(cloud.OOMError).WithDetail(msg).WithAbort()
+		cloud.SendTestRunEvents(cloudClient, k6.TestRunID(), log, events)
+		return true, nil
+	}
+
+	v1alpha1.UpdateCondition(k6, v1alpha1.TestRunRunning, metav1.ConditionFalse)
+	k6.GetStatus().Stage = "error"
+	_, err := r.UpdateStatus(ctx, k6, log)
+	return true, err
+}
+
+func checkRunnerOOMKilled(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *TestRunReconciler, cloudClient *cloudapi.Client) (bool, error) {
+	pods := &corev1.PodList{}
+	if err := r.List(ctx, pods, k6.ListOptions()); err != nil {
+		return false, err
+	}
+
+	return handleRunnerOOMKilled(ctx, log, k6, r, pods.Items, cloudClient)
 }
--- a/internal/controller/k6_finish.go
+++ b/internal/controller/k6_finish.go
@@ -20,6 +20,12 @@ func FinishJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *T
 	}
 
 	log.Info("Checking if all runner pods are finished")
+	if oomKilled, err := checkRunnerOOMKilled(ctx, log, k6, r, cloudClient); err != nil {
+		log.Error(err, "Could not check runner pods for OOM termination")
+		return
+	} else if oomKilled {
+		return
+	}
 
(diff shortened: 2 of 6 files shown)


internal/controller/common.go
```diff
@@ -192,3 +195,55 @@ func runTeardown(ctx context.Context, hostnames []string, log logr.Logger) {
...
+	msg := fmt.Sprintf("runner pod %s container %s was terminated with reason %s", podName, containerName, oomKilledReason)
+	log.Info(msg)
+
+	if isCloudTestRun(k6) {
+		if cloudClient == nil {
+			return true, errors.New("cloud client is not configured")
+		}
+		events := cloud.ErrorEvent(cloud.OOMError).WithDetail(msg).WithAbort()
+		cloud.SendTestRunEvents(cloudClient, k6.TestRunID(), log, events)
+		return true, nil
```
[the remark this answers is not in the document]
Guflly: Fixed in 83f2897. Successful delivery now transitions through StopJobs, and delivery errors are returned so reconciliation retries without advancing status.

internal/controller/common.go
```diff
@@ -192,3 +195,61 @@ func runTeardown(ctx context.Context, hostnames []string, log logr.Logger) {
 		log.Error(err, "Failed to invoke teardown()")
 	}
 }
+
+func isOOMKilled(status corev1.ContainerStatus) bool {
+	return status.State.Terminated != nil && status.State.Terminated.Reason == oomKilledReason ||
+		status.LastTerminationState.Terminated != nil && status.LastTerminationState.Terminated.Reason == oomKilledReason
+}
+
+func findOOMKilledContainer(pods []corev1.Pod) (string, string, bool) {
```
yorugac, line 204:
```suggestion
// on OOM, returns name of pod and name of OOM'ed container
func findOOMKilledContainer(pods []corev1.Pod) (string, string, bool) {
```

internal/controller/common.go
```diff
@@ -192,3 +195,61 @@ func runTeardown(ctx context.Context, hostnames []string, log logr.Logger) {
...
+			return true, nil
+		}
+		if cloudClient == nil {
+			return true, errors.New("cloud client is not configured")
+		}
+		events := cloud.ErrorEvent(cloud.OOMError).WithDetail(msg).WithAbort()
+		if err := cloud.SendTestRunEvents(cloudClient, k6.TestRunID(), log, events); err != nil {
+			return true, err
+		}
+		_, err := StopJobs(ctx, log, k6, r)
```
yorugac, line 238:
```suggestion
```
Cloud tests must wait until acknowledgment from the Cloud before aborting.

internal/controller/common.go
```diff
@@ -192,3 +195,61 @@ func runTeardown(ctx context.Context, hostnames []string, log logr.Logger) {
...
+		return true, err
+	}
+
+	v1alpha1.UpdateCondition(k6, v1alpha1.TestRunRunning, metav1.ConditionFalse)
+	k6.GetStatus().Stage = "error"
+	_, err := r.UpdateStatus(ctx, k6, log)
+	return true, err
```
yorugac, line 245:
```suggestion
	} else {
	  // this is OSS test run
	  v1alpha1.UpdateCondition(k6, v1alpha1.TestRunRunning, metav1.ConditionFalse)
  	k6.GetStatus().Stage = "error"
	_, err := r.UpdateStatus(ctx, k6, log)
	return true, err
	}
```
As mentioned in the [comment](https://github.com/grafana/k6-operator/issues/251#issuecomment-4083398027), OSS test runs can be moved to error state immediately, but not so with Cloud tests.

Also, it might be nice to kill pods on "error" state too... But that should be addressed separately (I'll make a note to think on this and create an issue / resurface one of the older ones).

internal/controller/common.go
```diff
@@ -25,6 +27,7 @@ import (
 
 const (
 	errMessageTooLong = "Creation of %s takes too long: your configuration might be off. Check if %v were created successfully."
+	oomKilledReason   = "OOMKilled"
```
yorugac, line 30: Let's move out all this new logic into `oom.go`, just as a unit test file is named.

internal/controller/testrun_controller.go
```diff
@@ -251,6 +251,11 @@ func (r *TestRunReconciler) reconcile(ctx context.Context, req ctrl.Request, log
 		}
 
 		if v1alpha1.IsTrue(k6, v1alpha1.CloudPLZTestRun) {
+			if oomKilled, err := checkRunnerOOMKilled(ctx, log, k6, r, cloudClient); err != nil {
```
yorugac, line 254:
```suggestion
            // OOM during main function of PLZ tests
			if oomKilled, err := checkRunnerOOMKilled(ctx, log, k6, r, cloudClient); err != nil {
```

internal/controller/testrun_controller.go
```diff
@@ -267,6 +272,11 @@ func (r *TestRunReconciler) reconcile(ctx context.Context, req ctrl.Request, log
 						return ctrl.Result{}, nil
 					}
 					runTeardown(ctx, hostnames, log)
+					if oomKilled, err := checkRunnerOOMKilled(ctx, log, k6, r, cloudClient); err != nil {
```
yorugac, line 275:
```suggestion
					// OOM during teardown
					if oomKilled, err := checkRunnerOOMKilled(ctx, log, k6, r, cloudClient); err != nil {
```

internal/controller/k6_start.go
```diff
@@ -95,14 +100,21 @@ func StartJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *Te
 	// setup
 
 	if v1alpha1.IsTrue(k6, v1alpha1.CloudPLZTestRun) {
-		if err, retry := runSetup(ctx, hostnames, log); err != nil {
+		setupErr, retry := runSetup(ctx, hostnames, log)
+		if oomKilled, err := checkRunnerOOMKilled(ctx, log, k6, r, cloudClient); err != nil {
```
yorugac, line 104:
```suggestion
		// OOM during setup function
		if oomKilled, err := checkRunnerOOMKilled(ctx, log, k6, r, cloudClient); err != nil {
```

internal/controller/k6_finish.go
```diff
@@ -20,6 +20,12 @@ func FinishJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *T
 	}
 
 	log.Info("Checking if all runner pods are finished")
+	if oomKilled, err := checkRunnerOOMKilled(ctx, log, k6, r, cloudClient); err != nil {
```
yorugac, line 23:
```suggestion
	// OOM for non-PLZ tests
	if oomKilled, err := checkRunnerOOMKilled(ctx, log, k6, r, cloudClient); err != nil {
```

internal/controller/oom_test.go
```diff
@@ -0,0 +1,222 @@
...
+			require.NoError(t, err)
+			require.True(t, found)
+			updated := &v1alpha1.TestRun{}
+			require.NoError(t, r.Get(context.Background(), client.ObjectKeyFromObject(k6), updated))
+			require.Equal(t, v1alpha1.Stage("error"), updated.Status.Stage)
+			require.True(t, v1alpha1.IsFalse(updated, v1alpha1.TestRunRunning))
+		})
+	}
+
+	t.Run("requests a cloud test run abort once", func(t *testing.T) {
```
yorugac, line 104:
```suggestion
	t.Run("requests a cloud test run abort at least once", func(t *testing.T) {
```
Just in case, that endpoint for SendTestRunEvents is supposed to be idempotent, so it's fine to call it more than once.

internal/controller/testrun_controller.go
yorugac: Also, please move this check after `UpdateCondition` / `UpdateStatus` call. If that fails, then no OOM check. If update is successful, then OOM check as now.

internal/controller/k6_start.go
```diff
@@ -95,14 +100,21 @@ func StartJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *Te
 	// setup
 
 	if v1alpha1.IsTrue(k6, v1alpha1.CloudPLZTestRun) {
-		if err, retry := runSetup(ctx, hostnames, log); err != nil {
+		setupErr, retry := runSetup(ctx, hostnames, log)
```
yorugac, line 103: Check for error from setup should be before OOM check. So OOM check is done only if setup returns OK.

internal/controller/oom_test.go
```diff
@@ -0,0 +1,222 @@
...
+		},
+		{
+			name: "restarted runner",
+			pod: corev1.Pod{
+				Status: corev1.PodStatus{ContainerStatuses: []corev1.ContainerStatus{{
+					Name:                 "k6",
+					LastTerminationState: corev1.ContainerState{Terminated: &corev1.ContainerStateTerminated{Reason: oomKilledReason}},
+				}}},
+			},
+			want: true,
```
yorugac, line 49: We don't actually support restarted runners 😅 I suggest to remove this test to avoid being misleading.
Related issue: #304

yorugac (changes requested):
Thank you for the PR on this, @Guflly! 
There are some things I noticed - most in the comments plus these:
- The linter is failing. 
- In `k6_finish.go` and `k6_start.go`, there is now 2 listings of Pods / Jobs happening. Is there a better way to do that?
- In this case, it's definitely worth it to check e2e tests as well + add an e2e test with OOM.

But TBH, looking at this implementation overall, I now wonder if we could actually have one OOM check closer to the beginning of reconciliation instead of 3-5 sprinkled around. The idea of checking in several places (from the issue thread) was because we might need to know in which part of k6 execution OOM happened exactly (for logging, for cloud events, for potential additional logic). But it _should_ be possible to recreate that from `TestRun` status:
| status | phase | ambiguous? |
| --- | --- | --- |
| `Stage == "created"` | during pod startup or setup | yes |
| `Stage == "started"` && `TeardownExecuted == False` | during main execution | no |
| `TeardownExecuted == True` or `Stage == "stopped"` | during/after teardown | no |

The first one we can improve in the future, and for now it's probably fine to leave it as just "setup" failure (initializer pod should catch the "pod startup" problem normally). 

WDYT about making this change? Of course, doing this would make some other comments invalid, likely 🙂 It should be done sometime after the `ShouldAbort` check, but before the stage switch, just in case.

----
Last but not least. FYI, we have a release next week: I'll probably be able to merge this only after that, even if the branch is ready. Since I also need to merge two more non-trivial PRs + test them, it's a bit of a timing issue: I just won't have time to test this PR sufficiently before the release. Hope it's alright (and yes, this branch will need a `git rebase` later, most likely).

yorugac:
Hi @Guflly, are you going to continue working on this one? :slightly_smiling_face:

Issue #757, opened 2026-03-31
TestRun deployments now broken


ts-sean-foley:
### Brief summary

We had working scripted TestRun deployments which send our test runs to EKS clusters in multiple regions, we pass env vars to the archive command with -e and we embed them in the TestRun with spec.runner.env declarations. I will provide more details as I get time, but what was working under 1.3.0  is broken in 1.3.1 with the initializer pods failing with an error: "accepts 1 arg(s), received 4"

### k6-operator version or image

1.3.1

### TestRun / PrivateLoadZone YAML

```apiVersion: k6.io/v1alpha1
kind: TestRun
metadata:
  name: poll-ts-sean-foley
  namespace: default
spec:
  parallelism: 1
  script:
    configMap:
      name: poll-ts-sean-foley
      file: poll.tar
  runner:
    image: chanfoli/k6-extended-influxdb:1.3
    env:
      - name: K6_AFF_CODE
        value: ts
      - name: K6_DNS
        value: ttl=30s,select=random,policy=onlyIPv4
      - name: K6_TEST_REGIONS
        value: docker-desktop
      - name: K6_LOG_LEVEL
        value: WARN
      - name: K6_TEST_TAG
        value: 2026-03-31 TS Dev Smoke 2
      - name: K6_TEST_SCALE
        value: dev
      - name: K6_TEST_TYPE
        value: smoke
      - name: K6_RES_SUM_URI
        value: https://REDACTED/main/results/insert
      - name: K6_INFLUXDB_BUCKET
        value: k6bucket
      - name: K6_INFLUXDB_ADDR
        value: http://telegraf-service:8086
      - name: K6_INFLUXDB_PUSH_INTERVAL
        value: 2s
      - name: K6_CONCURRENT_WRITES
        value: "12"
  arguments: --log-format json --out xk6-influxdb --tag aff_code=ts```

### Other environment details (if applicable)

_No response_

### Steps to reproduce the problem

Execute scripted TestRun deployment. 

### Expected behaviour

TestRun success

### Actual behaviour

Initializer fails with: accepts 1 arg(s), received 4

yorugac:
Thank you for the report, @ts-sean-foley. I've been looking at another similar report but without a full TestRun, so this is very helpful to see. 

[...]

PR #868, opened 2026-08-06
Add a `.spec.args` to TestRun CRD and switch PLZ to using it (2/3)

yorugac:
Supplementary work for https://github.com/grafana/k6-operator/issues/742 (not directly related to API changes but required to make arbitrary tags and env vars work).

Adding an alternative way to specify the k6 arguments. The initial approach, `.spec.arguments`, is ultimately faulty because it requires k6-operator to split the string into separate arguments. This is a complex task in a general case, prone to brittle results (see https://github.com/grafana/k6-operator/issues/162, https://github.com/grafana/k6-operator/issues/757); and given the Shell scripts in initializer job and in LocalFile test, it's also unsafe. The array of strings `.spec.args` allows a user to format their arguments as separate strings explicitly, which can then be passed to the k6 pods as is, minimizing the risk of misinterpretation of the values from k6-operator.

- New field in TestRun `.spec.args` as []string, which overrides `.spec.arguments` when present
- The initializer and LocalFile shell script revamped to support both forms, "simple" positional args for `.spec.args` and `sh -c` as before for joined `.spec.arguments` 
    - Since k6-operator avoided dragging in the full parsing of the string to avoid a complex implementation and to keep deps small, the only options are to offload it to the Shell (`.spec.arguments`) or to let user to be explicit about arguments as they know them best (`.spec.args`)
- PLZ flow switched to the `.spec.args`. In effect, this PR partially reverts bits from the PR https://github.com/grafana/k6-operator/pull/839: with `.spec.args`, the problem with arbitrary values in user env vars can be passed simply as "classic" `-e ...`. Future tags in PLZ will use the same approach (the next PR in this stack).

### DIFF
--- a/pkg/plz/worker.go
+++ b/pkg/plz/worker.go
@@ -161,8 +161,19 @@ func (w *PLZWorker) createTemplate(plz *v1alpha1.PrivateLoadZone) {
 	}
 }
 
-// the --log-output argument of PLZ runners; also used in unit tests
+// the --log-output argument of PLZ runners
 const plzLogOutputFormat = `--log-output=loki=https://cloudlogs.k6.io/api/v1/push,label.lz=%s,label.test_run_id=%s,header.Authorization=Token $(K6_CLOUD_TOKEN)`
+
+// plzk6Args builds the exact argv the worker is expected to produce.
+func plzk6Args(plzName string, testRunID string, trData *cloud.TestRunData) []string {
+	args := []string{"--out", "cloud"}
+	args = append(args, trData.TagArgs...)
+	args = append(args, "--no-thresholds")
+	args = append(args, fmt.Sprintf(plzLogOutputFormat, plzName, testRunID))
+	args = append(args, trData.EnvArgs...)
+	args = append(args, fmt.Sprintf("--include-system-env-vars=%t", trData.IncludeSystemEnvVars))
+	return args
+}
 
 // complete modifies tr with data from trData, which is specific for this test run.
 func (w *PLZWorker) complete(tr *v1alpha1.TestRun, trData *cloud.TestRunData) {
@@ -183,14 +194,7 @@ func (w *PLZWorker) complete(tr *v1alpha1.TestRun, trData *cloud.TestRunData) {
 	tr.Spec.TestRunID = trData.TestRunID()
 
 	// Building the argument list to k6.
-	args := []string{"--out", "cloud"}
-	args = append(args, trData.TagArgs...)
-	args = append(args, "--no-thresholds")
-	args = append(args, fmt.Sprintf(plzLogOutputFormat, w.plz.Name, trData.TestRunID()))
-	args = append(args, trData.EnvArgs...)
-	args = append(args, fmt.Sprintf("--include-system-env-vars=%t", trData.IncludeSystemEnvVars))
-
-	tr.Spec.Args = args
+	tr.Spec.Args = plzk6Args(w.plz.Name, trData.TestRunID(), trData)
 }
 
 // handle creates a new PLZ TestRun from the given test run id. The context is
--- a/pkg/plz/worker.go
+++ b/pkg/plz/worker.go
@@ -3,8 +3,6 @@ package plz
 import (
 	"context"
 	"fmt"
-	"slices"
-	"strings"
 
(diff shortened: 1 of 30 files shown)


pkg/plz/worker.go
```diff
@@ -234,6 +231,28 @@ func (w *PLZWorker) handle(ctx context.Context, testRunId string) {
...
+
+		events := cloud.ErrorEvent(cloud.K6OperatorStartError).
+			WithDetail("k6-operator: PLZ is misconfigured (outdated TestRun CRD); upgrade the k6-operator CRDs").
+			WithAbort()
+		cloud.SendTestRunEvents(w.poller.Client, testRunId, w.logger, events)
+
+		if err := w.k8sClient.Delete(ctx, tr); err != nil {
+			w.logger.Error(err, "Failed to delete the incomplete PLZ test run", "testRunId", testRunId)
+		}
+		return
```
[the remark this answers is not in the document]
yorugac: Possible. But in practice, worst case from the race: the initializer pod might indeed start and then be removed with the Delete, so impact should be pretty limited. For impact to be greater, GCk6 would have to be down or something.

This check is already a best-effort to help people avoid upgrade issues specific for this one release, so we don't need to track edge cases here.

pkg/types/k6cli.go
```diff
@@ -39,63 +45,42 @@ func ParseCLI(arguments string) (*CLI, error) {
...
-			continue
-		}
-		if args[i][0] == '-' {
-			end := lastArgV(i+1, args)
+	for i < len(argv) {
+		if len(argv[i]) > 0 && argv[i][0] == '-' {
+			end := lastArgV(i+1, argv)
 
-			if strings.HasPrefix(args[i], "--log-output") {
+			switch {
```
yorugac, line 53: Getting rid off mix of if and switch in favor of one unified switch.

pkg/plz/worker_test.go
```diff
@@ -69,6 +68,15 @@ func Test_complete_correctDefinitionOfTestRun(t *testing.T) {
 	// The following are the definitions that
 	// are expected from PLZ worker now.
 
+	// plzArgs builds the exact argv the worker is expected to produce.
+	plzArgs := func(plzName string, tagArgs []string, testRunID string, envArgs []string, includeSysEnvVars bool) []string {
```
neelisneel: Before this PR, the test compared against a hard-coded expected string. Now `plzArgs` builds the expected value using the same constant and the same order as complete() does, but if someone changes the argument order, the test changes with it and still passes. 

Maybe one case should keep a literal expected list?

neelisneel (dismissed):
LGTM, just 1 tiny comment regarding a test

pkg/plz/worker_test.go
yorugac: Yeah, I've wondered about that one too but not fully sure how best to approach it yet. 
Even with hard-coded string as before, if one changes arguments in `worker.go` without changing the string in test, the same thing happens: false positive of the unit test.

OTOH, the order here doesn't matter that much: what does matter is presence of the expected arguments. So maybe instead move `plzArgs` to `worker.go` to be used directly in `complete` too + add a test that checks `plzArgs` for presence of A, B, C instead?

yorugac: OK, added now so it won't get stuck in "TODO limbo" :smile: 
https://github.com/grafana/k6-operator/pull/868/changes/ccab42e008dc4dd946530bceae5388045dd275f7

pkg/types/k6cli.go
```diff
@@ -39,63 +45,42 @@ func ParseCLI(arguments string) (*CLI, error) {
...
-				strings.HasPrefix(args[i], "--user-agent") {
-				i = end
-				continue
-			}
+			case strings.HasPrefix(argv[i], "--block-hostnames"),
+				strings.HasPrefix(argv[i], "--blacklist-ip"),
+				strings.HasPrefix(argv[i], "--user-agent"):
+				// Unsupported by `k6 archive`.
 
-			switch args[i] {
```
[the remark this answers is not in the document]
yorugac: Yes, that's expected.

neelisneel (approved):

PR #876, opened 2026-08-17
fix: return error on stop / start job create failure

### LINKED ISSUE #840: bug: failed stop job creation leaves TestRun permanently stuck

AruneshDwivedi:
## Problem

When `StopJobs` fails to create the stop job (e.g. due to a transient API server error, quota issue, or RBAC problem), the error is logged but the function returns `ctrl.Result{}, nil` — telling controller-runtime "success, do not requeue".

Since the `TestRun` stage was never updated to `stopped`, nothing else triggers another reconcile, and the object is permanently stuck with no user-visible indication of why.

## Root Cause

In `internal/controller/k6_stop.go` lines 51-54:

```go
created, err := createJobIfNotExists(ctx, r.Client, stopJob)
if err != nil {
    log.Error(err, "Failed to launch k6 test stop job.")
    return res, nil   // res is zero value — no requeue, nil error
}
```

`res` is never assigned in this function, so it is always the zero `ctrl.Result{}`. Returning `nil` as the error signals success to controller-runtime. The reconciler is not retried, and the `TestRun` cannot progress.

## Expected Behaviour

A transient failure creating the stop job should be retried. The error should be returned so controller-runtime re-enqueues the reconcile request.

## Fix

Return `err` instead of `nil` on unexpected Create errors so the reconciler retries.

```go
if err = r.Create(ctx, stopJob); err != nil {
    if !apierrors.IsAlreadyExists(err) {
        log.Error(err, "Failed to launch k6 test stop job.")
        return res, err
    }
    log.Info("Stop job already exists, continuing")
}
```

Fixes #840

### DIFF
--- a/internal/controller/k6_start.go
+++ b/internal/controller/k6_start.go
@@ -121,7 +121,7 @@ func StartJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *Te
 	created, err := createJobIfNotExists(ctx, r.Client, starter)
 	if err != nil {
 		log.Error(err, "Failed to launch k6 test starter")
-		return res, nil
+		return res, err
 	}
 
 	if created {
--- a/internal/controller/k6_stop.go
+++ b/internal/controller/k6_stop.go
@@ -51,7 +51,7 @@ func StopJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *Tes
 	created, err := createJobIfNotExists(ctx, r.Client, stopJob)
 	if err != nil {
 		log.Error(err, "Failed to launch k6 test stop job.")
-		return res, nil
+		return res, err
 	}
 
 	if created {


yorugac (changes requested):
Thanks, @AruneshDwivedi. This seems like the right follow-up on #310 👍 The change is OK, but for consistency, could you please add the same change for `StartJobs` as well?

yorugac:
The change request in this PR wasn't addressed... Also, there's no need in a second PR in this case.

AruneshDwivedi:
Commit updated - added the same error return for `StartJobs` at line 124 for consistency. The starter job now also returns the error instead of swallowing it.

yorugac (approved):
Thanks for the update @AruneshDwivedi!

yorugac:
@AruneshDwivedi, the commits must be signed: merging is blocked without them. Could you please sign them?
image

AruneshDwivedi:
The signed-commits-info-action is blocking merge due to unsigned commits. To fix this, please rebase locally with signed commits:\n\n```bash\ngit checkout fix/stop-job-create-error\ngit rebase main\ngit commit --amend --no-edit --gpg-sign  # or use -S flag\n# Alternatively:\ngit commit --amend --reset-author --signoff\ngit push --force-with-lease origin fix/stop-job-create-error\n```\n\nOr you can squash and amend with a signed commit if GPG signing is preferred.

yorugac:
@AruneshDwivedi, commits here are still unsigned...

Issue #222, opened 2023-04-28
Implement timeout error


yorugac:
### Feature Description

Currently k6-operator does not have a concept of timeout: i.e. if runners for some reason stop responding, it is not being monitored. Implement a timeout period: if runners do not respond for this time period, the operator should enter `error` state.

yorugac:
Somewhat similarly to [OOM issue](https://github.com/grafana/k6-operator/issues/251), we need timeouts at 3 points: setup, main execution and teardown. 

An additional timeout that might make sense in a distributed execution: timeout for sending setup data -> evaluate how much it's needed in reality

## Problems

### Values for setup & teardown timeouts

k6 has defaults for timeouts: 60s, both for [setup](https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/#setup-timeout) and for [teardown](https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/#teardown-timeout), but they can be overridden in the script. For k6-operator to know the user-defined timeouts, initializer pod must be executed and parsed, which is no longer mandatory as of [1.3.0](https://github.com/grafana/k6-operator/releases/tag/v1.3.0). 

Instead of adding one more dependency on initializer pod, we can just use a larger default value for setup & teardown timeout, e.g. `10m`, and let k6 do the rest. 

_Open question: do we need a larger value and / or configuration for it?_

### Difference between PLZ and other types of tests

Sadly, we didn't yet implement support for run-once logic of setup / teardown in OSS nor in cloud output test runs, [issue](https://github.com/grafana/k6-operator/issues/223), which complicates this implementation.

With that in mind, we should add setup and teardown timeouts only for PLZ tests first, and then seek how to implement it for all test runs. As for the main exec timeout, ideally, it should be the same for all modes.

## Implementation

So far, we have implemented only one type of timeout: when initializer pod takes too long, it is logged and in case of Cloud test, aborted. The solution relies on conditions' timestamps, and it should be possible to extend to all kinds of timeouts. We might need new conditions for this.

PR #881, opened 2026-08-19
Fix duplicate PLZ setup execution across reconciles

### LINKED ISSUE #871: `Setup` can be executed twice for PLZ
For a PLZ test run, the k6 `setup` can be executed twice, depending on exact timing of reconciliation loops. That can happen because `runSetup` call doesn't have any conditional checks, so if the 2nd reconcile started before update of status and the move to "started" stage, it'll execute `setup` the second time.

How to detect that: run sufficient number of PLZ tests in a row in general. But using larger number of instances, longer `setup`  would increase probability.

To avoid that, add a condition for `SetupExecuted`, similar to `TeardownExecuted`. It must be updated to prior to running `setup` and added as a conditional check.

Non-PLZ test runs are not affected only because `setup` is skipped for them (#223).

jsk1004ha:
## What this changes

Fixes #871.

Adds a small `SetupExecuted` condition flow for PLZ test runs:

- `False` (or missing) is saved as `Unknown` before invoking `setup()`.
- A reconcile that observes `Unknown` waits instead of running setup or creating the starter.
- Retryable setup errors reset the condition to `False`.
- Successful setup changes it to `True`; later reconciles skip setup and continue the existing starter flow.

This keeps the change local to the existing condition and `StartJobs` paths. It does not add runner markers, claim IDs, timeouts, recovery phases, or setup-data APIs.

Validation: focused condition tests, `pkg/types` tests, controller package compilation, `go vet` for the changed packages, and `git diff --check`.

### DIFF
--- a/internal/controller/k6_start.go
+++ b/internal/controller/k6_start.go
@@ -95,18 +95,39 @@ func StartJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *Te
 	// setup
 
 	if v1alpha1.IsTrue(k6, v1alpha1.CloudPLZTestRun) {
-		if err, retry := runSetup(ctx, hostnames, log); err != nil {
-			if retry {
+		if _, exists := v1alpha1.LastUpdate(k6, v1alpha1.SetupExecuted); exists &&
+			v1alpha1.IsUnknown(k6, v1alpha1.SetupExecuted) {
+			return res, nil
+		}
+
+		if !v1alpha1.IsTrue(k6, v1alpha1.SetupExecuted) {
+			v1alpha1.UpdateCondition(k6, v1alpha1.SetupExecuted, metav1.ConditionUnknown)
+			if err = r.Status().Update(ctx, k6); err != nil {
 				return ctrl.Result{}, err
 			}
 
-			log.Error(err, "Setup function failed, requesting abort.")
-			events := cloud.ErrorEvent(cloud.SetupError).
-				WithDetail(fmt.Sprintf("setup function failed: %v", err)).
-				WithAbort()
-			cloud.SendTestRunEvents(cloudClient, k6.TestRunID(), log, events)
+			if err, retry := runSetup(ctx, hostnames, log); err != nil {
+				if retry {
+					v1alpha1.UpdateCondition(k6, v1alpha1.SetupExecuted, metav1.ConditionFalse)
+					if _, statusErr := r.UpdateStatus(ctx, k6, log); statusErr != nil {
+						return ctrl.Result{}, errors.Join(err, statusErr)
+					}
+					return ctrl.Result{}, err
+				}
 
-			return ctrl.Result{Requeue: false}, nil
+				log.Error(err, "Setup function failed, requesting abort.")
+				events := cloud.ErrorEvent(cloud.SetupError).
+					WithDetail(fmt.Sprintf("setup function failed: %v", err)).
+					WithAbort()
+				cloud.SendTestRunEvents(cloudClient, k6.TestRunID(), log, events)
+
+				return ctrl.Result{Requeue: false}, nil
+			}
+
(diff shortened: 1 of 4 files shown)


internal/controller/k6_start.go
```diff
@@ -62,18 +96,15 @@ func StartJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *Te
...
-				}
+		} else if time.Since(t).Minutes() > 5 {
+			msg := fmt.Sprintf(errMessageTooLong, "runner pods", "runner jobs and pods")
+			log.Info(msg)
+
+			if v1alpha1.IsTrue(k6, v1alpha1.CloudTestRun) {
+				events := cloud.ErrorEvent(cloud.K6OperatorStartError).
+					WithDetail(msg).
+					WithAbort()
+				cloud.SendTestRunEvents(cloudClient, k6.TestRunID(), log, events)
```
yorugac: Please keep original comments and formatting. This part wasn't supposed to change at all

yorugac (changes requested):
@jsk1004ha, thanks for the PR, but I'm afraid this approach cannot be merged.

The initial bug issue describes a narrow fix, while this PR adds much more complex logic in comparison, with a whole new state machine attached.

If you'd like to continue working on this, please narrow this down significantly. It is not the goal right now to have such complexity around `setup`. 
I’m not leaving much comments because TBH, this PR needs a substantial revamp. For example, keep the condition and add just a simple save & check flow - that should be a starting point. Everything else is up to discussion and e2e testing.

Also, was an LLM used to generate this PR? It looks so to me. This is not prohibited AFAIK, but my personal request is to keep the generation "contained" and well reviewed, and the changes on a smaller side. If something is unclear, it's also best to ask 🙂

internal/controller/k6_start.go
jsk1004ha: @yorugac Oh, I apologize. It seems I gradually lost focus while working with the AI ​​to fix the remaining issues after the initial bug fixes. I will ensure more thorough PR reviews in the future. I’ll proceed with the corrections based on your feedback!

jsk1004ha:
Fixed — I restored the original comments and formatting in this block. I also rewrote the PR around a small SetupExecuted save/check flow and removed the unrelated changes and the larger state machine. Thanks for the feedback!

internal/controller/k6_start.go
```diff
@@ -95,18 +95,39 @@ func StartJobs(ctx context.Context, log logr.Logger, k6 *v1alpha1.TestRun, r *Te
...
-		if err, retry := runSetup(ctx, hostnames, log); err != nil {
-			if retry {
+		if _, exists := v1alpha1.LastUpdate(k6, v1alpha1.SetupExecuted); exists &&
+			v1alpha1.IsUnknown(k6, v1alpha1.SetupExecuted) {
+			return res, nil
+		}
+
+		if !v1alpha1.IsTrue(k6, v1alpha1.SetupExecuted) {
+			v1alpha1.UpdateCondition(k6, v1alpha1.SetupExecuted, metav1.ConditionUnknown)
+			if err = r.Status().Update(ctx, k6); err != nil {
```
yorugac, line 105: A note for the record. This logic (start of conditional checks) will likely fail during restarts of k6-operator: it won't know to check for time and TestRun can get stuck when the restart happens in-between of this whole `setup` block. We'll be basically fixing this separately, in #222.

yorugac:
cursor review

yorugac (approved):
@jsk1004ha, thank you for updating the PR. It's much more readable in one go now 😄 

PS I've left a comment but that's "for the record": it's not meant as a change request here.

yorugac:
@jsk1004ha, it appears that the signatures are absent: it's a requirement for merging  in Grafana. Can you please sign the commits?
image

jsk1004ha:
@yorugac The GitHub signature has been verified.

yorugac (approved):
Thanks for the update @jsk1004ha! 🙌 
I should be able to merge it now...

Issue #186, opened 2023-02-02
Networking improvements


yorugac:
This is an umbrella issue to improve networking in the operator. Currently, details are best described here: https://github.com/grafana/k6-operator/pull/146#issuecomment-1351766577

And also PR for IPv6: https://github.com/grafana/k6-operator/issues/171

It's best to address these points together so that different use cases do not conflict with each other.

yorugac:
PR https://github.com/grafana/k6-operator/pull/882 prevents runner readiness probes from indefinitely blocking reconciliation and formats IPv6 ClusterIPs correctly. Timeout of HTTP check is hard-coded to 1 sec: enough to get the value under normal circumstances, not too much to get stuck.

As a follow-up to it, it might make sense to make the check of services concurrent instead of sequential. (If someone picks it up, mind  that there is no limit on parallelism value (up to user) so it will have to be batched probably.)

Is there any other use case or tricky networking setup we need to cover?

PR #882, opened 2026-08-19
fix: bound and harden service readiness probes

jsk1004ha:
## What changed

- propagate the reconcile context into each runner service readiness request
- cap the `/v1/status` request at two seconds so an unresponsive endpoint cannot stall reconciliation indefinitely
- construct endpoints with `net.JoinHostPort`, including bracketed IPv6 ClusterIPs
- preserve the existing readiness rule that HTTP responses below 400 are considered ready
- simplify readiness probing and cover the actual `200 OK` and `500 Internal Server Error` responses with `httptest.Server`
- retain focused regression coverage for request timeouts and parent cancellation

## Why

The controller previously used bare `http.Get` with no timeout and interpolated the ClusterIP directly into the URL. A service that accepted a connection but never responded could block the controller indefinitely, and an IPv6 ClusterIP produced an invalid URL such as `http://2001:db8::1:6565/...`.

This is proactive hardening related to the broader networking and IPv6 tracking issues (#186 and #171); there was no dedicated issue or competing PR for this specific probe path when the change was prepared.

## Validation

```text
gofmt (changed Go files)
git diff --check
go test ./internal/controller -run 'TestProbeServiceStatus' -count=20
go test ./internal/controller -count=1
go test $(go list ./... | grep -v /e2e) -count=1
go vet $(go list ./... | grep -v /e2e)
```

### DIFF
--- a/internal/controller/service_ready.go
+++ b/internal/controller/service_ready.go
@@ -1,0 +1,65 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	serviceStatusPort           = "6565"
+	serviceStatusRequestTimeout = 2 * time.Second
+)
+
+type httpDoer interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
+func serviceStatusEndpoint(clusterIP string) (string, error) {
+	if net.ParseIP(clusterIP) == nil {
+		return "", fmt.Errorf("invalid service ClusterIP %q", clusterIP)
+	}
+
+	return fmt.Sprintf("http://%s/v1/status", net.JoinHostPort(clusterIP, serviceStatusPort)), nil
+}
+
+func isServiceReady(ctx context.Context, log logr.Logger, service *v1.Service) bool {
+	endpoint, err := serviceStatusEndpoint(service.Spec.ClusterIP)
+	if err != nil {
+		log.Error(err, "Failed to build service status endpoint", "service", service.Name)
+		return false
+	}
+
+	return probeServiceStatus(ctx, log, http.DefaultClient, endpoint, serviceStatusRequestTimeout)
+}
+
+func probeServiceStatus(
+	ctx context.Context,
+	log logr.Logger,
+	client httpDoer,
+	endpoint string,
+	timeout time.Duration,
+) bool {
+	requestCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(requestCtx, http.MethodGet, endpoint, nil)
+	if err != nil {
+		log.Error(err, "Failed to build service status request", "endpoint", endpoint)
+		return false
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		log.Error(err, "Failed to get service status", "endpoint", endpoint)
+		return false
+	}
+	defer resp.Body.Close() //nolint:errcheck
+
+	return resp.StatusCode < http.StatusBadRequest
+}
--- a/internal/controller/service_ready.go
+++ b/internal/controller/service_ready.go
@@ -2,7 +2,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 	"net"
 	"net/http"
 	"time"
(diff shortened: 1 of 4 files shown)


internal/controller/service_ready.go
```diff
@@ -0,0 +1,65 @@
...
+	"net/http"
+	"time"
+
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	serviceStatusPort           = "6565"
+	serviceStatusRequestTimeout = 2 * time.Second
```
yorugac: How was this number decided?

internal/controller/service_ready.go
```diff
@@ -0,0 +1,65 @@
...
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/go-logr/logr"
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	serviceStatusPort           = "6565"
```
yorugac:
```suggestion
	defaultServicePort           = "6565"
```

internal/controller/service_ready.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type httpDoer interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
+func serviceStatusEndpoint(clusterIP string) (string, error) {
+	if net.ParseIP(clusterIP) == nil {
+		return "", fmt.Errorf("invalid service ClusterIP %q", clusterIP)
+	}
```
yorugac: In which case can this path happen?

I suspect this `serviceStatusEndpoint` can be removed as redundant and then unit test can be simplified. Answering this question precisely should help with that.

internal/controller/service_ready_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	tests := []struct {
+		name       string
+		statusCode int
+		wantReady  bool
+	}{
+		{name: "success", statusCode: http.StatusNoContent, wantReady: true},
+		{name: "redirect", statusCode: http.StatusTemporaryRedirect, wantReady: true},
+		{name: "client error", statusCode: http.StatusBadRequest},
+		{name: "server error", statusCode: http.StatusServiceUnavailable},
```
yorugac: Were these matched against what can actually be returned there?

yorugac (changes requested):
Thanks for the PR, @jsk1004ha! I like the idea of adding timeouts and `JoinHostPort`, but I request some changes in logic: see the comments. 

Also, please try to simplify the unit tests: I think `httptest.Server` would be a better approach and arguably more readable, instead of `httpDoer`. I also suspect that some of those tests cases can be removed as redundant: this needs some thinking (more free time than I have ATM, but please think on that).

internal/controller/service_ready.go
jsk1004ha: I set a conservative upper limit of 2 seconds for cluster internal status requests, but this wasn't based on existing project settings or observed behavior. As you pointed out, that value was indeed chosen arbitrarily. Could you perhaps suggest a better way to determine this figure?

jsk1004ha: Alternatively, since StartJobs already readjusts at one-second intervals, should I set the timeout to one second as well?

internal/controller/service_ready.go
jsk1004ha: This path should not occur for runner Services created by the operator. NewRunnerService creates a regular ClusterIP Service without setting clusterIP, so Kubernetes allocates a valid address, and the operator does not create these as headless Services. An explicitly invalid ClusterIP would be rejected by the API server.  
I agree that the validation and serviceStatusEndpoint helper are redundant here. I'll remove them and construct the address directly with net.JoinHostPort.

internal/controller/service_ready_test.go
jsk1004ha: I’m very sorry. These examples were not selected to compare with the responses returned by the k6 status endpoint, but rather to test the existing < 400 conditions. If you look at the handler, GET /v1/status returns a 200 OK on success and a 500 Internal Server Error in the event of an encoding error. The 204, 307, and 400 examples aren’t representative. 😥
I think it would be best to replace these examples with meaningful success and failure cases. I’ll also use `httptest.Server` to streamline the tests.

Translated with [DeepL.com](https://www.deepl.com/?utm_campaign=product&utm_source=web_translator&utm_medium=web&utm_content=copy_free_translation) (free version)

internal/controller/service_ready.go
jsk1004ha: Implemented in `8c0e51e`: renamed the constant to `defaultServicePort`.

internal/controller/service_ready.go
jsk1004ha: Implemented in `8c0e51e`. I removed the redundant ClusterIP validation and `serviceStatusEndpoint` helper, and now construct the endpoint directly with `net.JoinHostPort`.

internal/controller/service_ready_test.go
jsk1004ha: Implemented in `8c0e51e`. The tests now use `httptest.Server` instead of the custom `httpDoer`/`RoundTripper`, and the status cases are limited to the actual `200 OK` and `500 Internal Server Error` responses. Timeout and parent-cancellation coverage remain in place.

jsk1004ha:
Addressed the requested changes in `8c0e51e`:

- renamed the port constant to `defaultServicePort`
- removed the redundant ClusterIP validation and endpoint helper
- replaced the custom HTTP test double with `httptest.Server`
- limited status cases to `200 OK` and `500 Internal Server Error`
- kept the two-second production timeout and the timeout/parent-cancellation coverage

Validation completed with the focused controller tests, the full non-e2e test suite on Linux, `go vet`, and `git diff --check`.

internal/controller/service_ready.go
yorugac: I see, thanks.
🤔 It's actually quite difficult to be certain of the number here, IMO. But it likely shouldn't be large: it'd block the controller. The runner pods are already running here, so this error is mostly about k6 process taking a moment to start up fully (milliseconds with healthy image) or some network error which might also be transient. Either way, the failure will be caught on the error reconcile and seen in the logs - there's no need to block it on HTTP timeout. 
So for these reasons, I'd lower the default to 1 sec. It _might_ lead to false negatives from this check sometimes, but that's only a delay of the start until the next reconciliation in practice (on healthy cluster).

yorugac (commented):
Thanks for the update, @jsk1004ha! I think the PR is ready for merge: please just change the default to 1 sec (I've explained the reasoning in the comment). And again, the commits must be all signed 🙂

jsk1004ha:
Addressed the remaining items in `d76302b`:

- lowered `serviceStatusRequestTimeout` from 2 seconds to 1 second
- rewrote the second commit with a valid GitHub-verified signature

Both commits in the PR now show as verified. The new commit differs from the previous revision only by the requested timeout change. CI has been triggered again; requesting re-review once it completes.

jsk1004ha:
The new fork commit requires maintainer approval before the GitHub Actions workflows can run (`action_required`). Could a maintainer please approve the workflow runs and re-review the timeout/signature update? The signed-commits check and CLA already pass.

yorugac (approved):
@jsk1004ha, thank you! 🙌

PR #884, opened 2026-08-22
fix(controller): return error on start job create failure

AruneshDwivedi:
Fixes same issue as #876 for the StartJobs path. When creating the starter job fails, the error was swallowed and the reconciler returned success, leaving the TestRun permanently stuck.


yorugac:
This is a partial duplicate of https://github.com/grafana/k6-operator/pull/876
That PR awaits your action @AruneshDwivedi, but there is no reason to have two PRs here. So closing this one.

PR #892, opened 2026-09-01
fix(plz): prevent setup() from running twice on PLZ test runs

### LINKED ISSUE #871: `Setup` can be executed twice for PLZ

AruneshDwivedi:
The runSetup() call in StartJobs had no guard, so concurrent
reconcile loops could execute setup() multiple times before the
status updated to 'started'.

Add a SetupExecuted condition (mirroring the existing TeardownExecuted
pattern) and check it before calling runSetup(). Set the condition
to true after successful setup completion.

fixes #871

### DIFF
--- a/api/v1alpha1/k6conditions.go
+++ b/api/v1alpha1/k6conditions.go
@@ -20,6 +20,10 @@ const (
 	// This condition can be used only in PLZ test runs.
 	TeardownExecuted = "TeardownExecuted"
 
+	// SetupExecuted indicates whether the `setup()` has been executed on one of the runners.
+	// This condition can be used only in PLZ test runs.
+	SetupExecuted = "SetupExecuted"
+
 	// CloudTestRun indicates if this test run is supposed to be a cloud test run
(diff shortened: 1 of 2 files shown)


yorugac:
The issue https://github.com/grafana/k6-operator/issues/871 was actually already fixed with an earlier PR from another contributor. Nevertheless, thank you for the effort, @AruneshDwivedi!

PR #903, opened 2026-09-15
feat: respect K6_LOG_OUTPUT from GCk6 for PLZ log output

fornfrey:
The Loki endpoint of PLZ runners was hard-coded, so it couldn't be pointed at dev environment and did not respect the custom headers and labels defined on the backend.

The base value of `--log-output` now comes from `gck6_env_vars.K6_LOG_OUTPUT`, falling back to `https://cloudlogs.k6.io/api/v1/push`. 

`label.lz`, `label.test_run_id` and the auth header are still appended by k6-operator.


pkg/cloud/apitypes.go
```diff
@@ -206,6 +219,18 @@ func (lz *LZConfig) reservedEnvVars() []corev1.EnvVar {
 	return ev
 }
 
+// LogOutput returns the base value of the k6 `--log-output` option for this
+// test run: it is either the one defined by GCk6 or DefaultLogOutput.
+func (lz *LZConfig) LogOutput() string {
+	// Trailing commas would produce an empty parameter for k6 to choke on,
```
yorugac, line 225: Why would there _be_ stray commas in the first place 😂

pkg/cloud/apitypes.go
```diff
@@ -11,10 +11,16 @@ import (
 
 // GCk6 can set only a limited number of env vars to k6 process:
 // these are known and whitelisted with this "const" map.
+//
+// Note: K6_LOG_OUTPUT is deliberately not here. It is consumed by LogOutput()
+// and passed to the runners as a part of the `--log-output` argument instead,
+// because k6-operator has to amend its value with the labels and the auth
+// header. Setting it as an env var as well would be redundant, as the CLI
+// option takes precedence, and it would only invite drift between the two.
```
yorugac, line 19: 🤔 I vaguely remember the idea to move log output from the CLI option to an env var... But yeah, I suppose until it happens, this is more or less correct.

yorugac (approved):
Thanks for the fix, @fornfrey 🙌

